PR #8, opened 2016-04-07
Fix WES-21 Cannot rename a company

x4d3:
fix the bug in "her" gem by continuing the patching.
The error is the original gem is due to the next call that does not returns the hash
that hash is then given in the inject method as first argument, and since it is nil, cause an error


x4d3:
@ouranos could you please review

PR #9, opened 2016-04-08
fix issue with company renaming

x4d3:
fix the bug in "her" gem by continuing the patching.
The error is the original gem is due to the next call that does not returns the hash
that hash is then given in the inject method as first argument, and since it is nil, cause an error

### DIFF
--- a/core/lib/her_extension/model/parse.rb
+++ b/core/lib/her_extension/model/parse.rb
@@ -33,7 +33,22 @@ module Her
             filtered_attributes
           end
         end
-        
+
+        # @private
+        # TODO: Handle has_one
+        def embeded_params(attributes)
+          associations[:has_many].select { |a| attributes.include?(a[:data_key])}.compact.inject({}) do |hash, association|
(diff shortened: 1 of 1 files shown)


ouranos:
Good catch.

It'd be nice to make more obvious what was patched as it's hard to see without looking at the source. It's the usual problem with monkey patching as we often copy/paste the original implementation.

Maybe something like:

```ruby
        # @private
        # TODO: Handle has_one
        def embeded_params(attributes)
          associations[:has_many].select { |a| attributes.include?(a[:data_key])}.compact.inject({}) do |hash, association|
            params = attributes[association[:data_key]].map(&:to_params)
            # PATCH - Return hash 
            next hash if params.empty?
            # EO PATCH
            if association[:class_name].constantize.include_root_in_json?
              root = association[:class_name].constantize.root_element
              hash[association[:data_key]] = params.map { |n| n[root] }
            else
              hash[association[:data_key]] = params
            end
            hash
          end
```

What do you think?

x4d3:
I agree,  I'll update.

PR #10, opened 2016-04-12
[MNOE-64] Fix bug introduced in 552c0b7b

ouranos:


ouranos:
@x4d3 FYI

ouranos:
See 552c0b7b

PR #15, opened 2016-04-22
Admin Platform: User setup workflow

ouranos:


ouranos:
@BrunoChauvet Code Review please?

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -36,5 +36,86 @@ def count
...
+        user = MnoEnterprise::User.new(user_params)
+        user.skip_confirmation_notification!
+        user.save
+
+        # Reset the confirmation field so we can track when the invite is send - #confirmation_sent_at is when the confirmation_token was generated (not sent)
+        # Not ideal as we do 2 saves, and the previous save trigger a call to the backend to validate the token uniqueness
+        user.assign_attributes(confirmation_sent_at: nil, confirmation_token: nil)
+        user.save
+        user
+      end
```
BrunoChauvet: Would this work?

```ruby
user = MnoEnterprise::User.new(user_params)
user.skip_confirmation_notification!
user.confirmation_sent_at = nil
user.confirmation_token = nil
user.save
user
```

api/spec/controllers/mno_enterprise/jpi/v1/app_instances_sync_controller_spec.rb
```diff
@@ -77,18 +77,18 @@ module MnoEnterprise
     end
 
     describe "POST #create" do
-      xit "to spec: cannot stub 'post /app_instances_syncs data%5Bmode%5D=a_mode'"
+      it "to spec: cannot stub 'post /app_instances_syncs data%5Bmode%5D=a_mode'"
```
BrunoChauvet: should we keep this spec?

core/spec/models/mno_enterprise/ability_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+# TODO: add more ability tests
+RSpec.describe MnoEnterprise::Ability, type: :model do
+  subject(:ability) { described_class.new(user) }
+  let(:user) { FactoryGirl.build(:user, admin_role: admin_role) }
+  let(:admin_role) { nil }
+
+  context 'when User#admin_role is admin' do
+    let(:admin_role) { 'admin' }
+    it { is_expected.to be_able_to(:manage_app_instances, MnoEnterprise::Organization) }
+  end
```
BrunoChauvet: Should probably test the negative scenario as well

api/spec/controllers/mno_enterprise/jpi/v1/app_instances_sync_controller_spec.rb
ouranos: I left it to still get a pending warning as the spec below is barely testing anything (just the authorization)

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
ouranos: From the devise confirmable model:
```ruby
before_create :generate_confirmation_token, if: :confirmation_required?
after_create  :send_on_create_confirmation_instructions, if: :send_confirmation_notification?
```

`.skip_confirmation_notification!` only skip the 2nd callback and not the first one so I have to reset the fields.

I could monkey patch devise `#generate_confirmation_token` to exit if some flag is set if you think that's better.

BrunoChauvet: Leave it like this if it is too much pain, we can have 2 API calls in this scenario.

PR #18, opened 2016-04-28
fix caching issue

x4d3:


ouranos:
Thanks!

Issue was introduced in `3.0` branch and isaacs/github#18 is not fixed so closing and reopening a new PR.

PR #30, opened 2016-05-16
Add OrganizationAppInstance in the model to be able to provision instance using API

x4d3:


x4d3:
@ouranos could you please review ?

PR #34, opened 2016-05-26
Use price instead of price_tag

x4d3:
I think price should be send as a hash instead of a formatted string (because some information is lost, for example $12 can be 12  USD or 12 AUD)
See PR: https://github.com/maestrano/Maestrano/pull/1594

### DIFF
--- a/core/app/pdf/mno_enterprise/invoice_pdf.rb
+++ b/core/app/pdf/mno_enterprise/invoice_pdf.rb
@@ -85,7 +85,7 @@ module MnoEnterprise
       price = item[:price]
       if price
         # Money hash are automatically parsed to Money in core/lib/her_extension/middleware/mnoe_api_v1_parse_json.rb
-        price.format
+        money(price)
       else
         item[:price_tag]
       end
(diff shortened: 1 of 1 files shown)


x4d3:
@ouranos Could you please code review ?

core/app/pdf/mno_enterprise/invoice_pdf.rb
```diff
@@ -116,6 +127,12 @@ def image_path(name)
       File.exists?(app_path) ? app_path : engine_path
     end
 
+    # Format a money object
+    def money(m)
+      "#{m.format(symbol: false)} #{m.currency_as_string}"
+    end
```
ouranos: Why not just `m.format`? Is it to have the currency after the amount?

x4d3: m.format is giving something like $ 8, that is ambiguous.

ouranos:
Happy with this as long as @alachaum is happy with maestrano/Maestrano#1594 :)

core/app/pdf/mno_enterprise/invoice_pdf.rb
ouranos: Ah got it, `currency_as_string` will give USD / AUD

PR #38, opened 2016-06-24
Correct root redirection in template routes

x4d3:


x4d3:
@ouranos Could you please review ?

PR #47, opened 2016-07-06
[MNOE-97] Add missing value in rails template

winnietan:


rails-template/mnoe-app-template.rb
```diff
@@ -70,7 +70,7 @@ def gemfile_entries
 end
 
 task :test do
-  Rake::Task['test:all'].invoke
+  Rake::Task['test:all'].invokecd
```
ouranos: Typo?  ;)

rails-template/mnoe-app-template.rb
```diff
@@ -81,6 +81,17 @@ def gemfile_entries
 # Create uat environment
 copy_file File.join(destination_root, 'config/environments/production.rb'), 'config/environments/uat.rb'
 
+# Edit config/environments/*.rb
+Dir["config/environments/*.rb"].each do |file|
+  insert_into_file file, "\n  config.action_mailer.default_url_options = {host: 'localhost:7000'} \n", before: /^end$/
```
ouranos: Remove the trailing space between "{host: 'localhost:7000'}" and "\n"

PR #48, opened 2016-07-07
[MNOE-56] Google Tag Manager Container

claire00:


ouranos:
@claire00 All good but your editor is removing the new line at end of file (see the red sign on the diff)
Did you get the editorconfig and the sublime plugin as explained on the onboarding wiki?

For more details you can read this article: https://robots.thoughtbot.com/no-newline-at-end-of-file

PR #52, opened 2016-07-18
[MNOE-109] Better App title management

winnietan:


core/lib/generators/mno_enterprise/install/templates/stylesheets/variables.less
```diff
@@ -29,6 +29,7 @@
 /*                              Login Page                               */
 /*-----------------------------------------------------------------------*/
 @login-bg-color:                     @bg-main-color;
+@login-bg-img:						 "";
```
ouranos: Use spaces instead of tab.

PR #61, opened 2016-08-12
[MNOE-125] Fix filtering on empty array

ouranos:
If a filter is an empty array, it'll be excluded when calling
`params.to_query`, so convert it to nil instead.

eg: where('id.in' => []) => {'id.in' => nil}

### DIFF
--- a/core/lib/her_extension/model/relation.rb
+++ b/core/lib/her_extension/model/relation.rb
@@ -26,6 +26,9 @@ module Her
       # Use filter instead of raw parameters
       def where(params = {})
         return self if !params || params.empty?
+        # If a value is an empty array, it'll be excluded when calling params.to_query, so convert it to nil instead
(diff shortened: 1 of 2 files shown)


ouranos:
@alachaum @BrunoChauvet what do you think of this? Any unintended side effects?

BrunoChauvet:
Should we rather interpret this as an invalid query and raise a 400 error? I would at the very least log a warning when this occurs.

ouranos:
@BrunoChauvet good idea, see my last commit

@alachaum thoughts? Is there any case where we'd need `.where('xxx' => [])` ?

alachaum:
do we really need to raise an error? What's the active record behaviour when you do User.where(email: [])

ouranos:
Removed last commit and reverted to filtering on `nil` (same behavior as ActiveRecord)

PR #65, opened 2016-08-19
[MNOE-117] Generate stub font-faces.less

ouranos:
Since `main.less` imports `font-faces.less`, creates this file when
generating a new project.


alexnoox:
LGTM

PR #66, opened 2016-08-22
Admin Panel bug fixes

alexnoox:


alexnoox:
@ouranos I always needed to do this before building the admin:
```
rm -rf public/admin
rm -rf tmp/build/admin
```
The last commit solves that ;)

alexnoox:
@ouranos The last commit fix the exit menu redirection. It also fixes the admin_role display on the user view.

ouranos:
Good. Merging on the 3.1 branch

PR #70, opened 2016-08-26
[MNOE-137] Add Her middleware to handle errors

ouranos:
Handle connection and server errors

Mnoe api call will now raise `Faraday::Error::ConnectionFailed` or `Faraday::Error::ClientError` instead of returning `nil`

When the call is valid but the resource is not found it'll still return `nil` to keep the same behavior.


ouranos:
@alachaum FYI

BrunoChauvet:
LGTM

PR #75, opened 2016-09-07
[MNOE-107] Fix admin finance page

ouranos:
Tenant#current_billing_amount to be calculated via mnohub for performance reason (and also to
perform the currency conversion)

Admin Finance page:
- Fix invoices table amount
- Add extra finance KPIs

### DIFF
--- a/api/app/controllers/mno_enterprise/jpi/v1/admin/invoices_controller.rb
+++ b/api/app/controllers/mno_enterprise/jpi/v1/admin/invoices_controller.rb
@@ -13,35 +13,49 @@ module MnoEnterprise
 
     # GET /mnoe/jpi/v1/admin/invoices/current_billing_amount
     def current_billing_amount
-      data = Rails.cache.fetch('tenant_admin/current_billing_amount', expires_in: ADMIN_CACHE_DURATION) do
-        billing = MnoEnterprise::Organization.all.map(&:current_billing).sum(Money.new(0))
-        {current_billing_amount: {amount: billing.amount, currency: billing.currency_as_string}}
-      end
-      render json: data
+      # Backward compatibility with old MnoHub (<= v1.0.2)
+      # TODO: Remove once all mnohub are migrated to newer versions
+      tenant.respond_to?(:current_billing_amount) && current_billing = tenant.current_billing_amount
+
+      render json: {current_billing_amount: format_money(current_billing)}
     end
 
     # GET /mnoe/jpi/v1/admin/invoices/last_invoicing_amount
     def last_invoicing_amount
-      tenant_billing = MnoEnterprise::Tenant.get('tenant').last_customers_invoicing_amount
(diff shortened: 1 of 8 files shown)


api/app/controllers/mno_enterprise/jpi/v1/admin/invoices_controller.rb
```diff
@@ -13,11 +13,18 @@ def show
 
     # GET /mnoe/jpi/v1/admin/invoices/current_billing_amount
     def current_billing_amount
-      data = Rails.cache.fetch('tenant_admin/current_billing_amount', expires_in: ADMIN_CACHE_DURATION) do
-        billing = MnoEnterprise::Organization.all.map(&:current_billing).sum(Money.new(0))
-        {current_billing_amount: {amount: billing.amount, currency: billing.currency_as_string}}
-      end
-      render json: data
+      tenant = MnoEnterprise::Tenant.get('tenant')
```
BrunoChauvet: To get a singular resource, do we need to specify a fake id? Could it be `MnoEnterprise::Tenant.get` ?

api/app/controllers/mno_enterprise/jpi/v1/admin/invoices_controller.rb
```diff
@@ -13,11 +13,18 @@ def show
...
+
+      # Backward compatibility with old MnoHub -> to remove later
```
BrunoChauvet: Probably add the tag `TODO` here

BrunoChauvet:
I would just add a `TODO` comment with the required version of mnohub so it's easier to do some clean up (don't forget to add the same todo to the old_tenant factory)

api/app/controllers/mno_enterprise/jpi/v1/admin/invoices_controller.rb
ouranos: It's not a fake id actually. It's the resource name:
```
>> MnoEnterprise::Tenant.get('tenant')
=> GET /api/mnoe/v1/tenant
```

I'll check if I can find something cleaner

ouranos: @BrunoChauvet @x4d3 
How about:
```ruby
module MnoEnterprise
  class Tenant < BaseResource
    def self.show
      self.get('tenant')
    end
  end
end
```
So we can use `MnoEnterprise::Tenant.show` in the controller? Or I could override `get`?

BrunoChauvet: That looks cleaner indeed :+1:

ouranos:
@BrunoChauvet see last commit

BrunoChauvet:
Looks good

ouranos:
Squashed

PR #76, opened 2016-09-08
[MNOE-148] Only display connectors in admin onboarding

ouranos:
Admin Panel - Only display connectors application in the "Connect your Apps" step of the onboarding.


BrunoChauvet:
LGTM

PR #78, opened 2016-09-09
[MNOE-149] Allow impersonate_redirect_uri

ouranos:
Impersonate now takes a redirect_uri parameters which will be the url
the admin is redirected to when exiting impersonation.


BrunoChauvet:
:+1:

PR #80, opened 2016-09-11
[MNOE-149] Use redirect_uri when impersonating uri

ouranos:
Admin will be redirected to the original page when exiting impersonation


BrunoChauvet:
:+1:

PR #82, opened 2016-09-15
Add Intercom Server Side

x4d3:
- Separated event logger in 2 listener: Audit Events Listener and Intercom Events Listener


api/lib/mno_enterprise/event_logger.rb
```diff
@@ -2,23 +2,15 @@
...
-              key: key,
-              user_id: current_user_id,
-              description: description,
-              metadata: format_metadata(metadata, object),
-              subject_type: object.class.name,
-              subject_id: object.id
-          }})
-    rescue Net::ReadTimeout
-      # Meant to fail
+      formated_metadata = format_metadata(metadata, object)
```
ouranos: `formatted`? :)

api/lib/mno_enterprise/event_logger.rb
```diff
@@ -2,23 +2,15 @@
 
 module MnoEnterprise
   class EventLogger
-    include HTTParty
-    base_uri "#{MnoEnterprise.mno_api_private_host || MnoEnterprise.mno_api_host}/api/mnoe/v1/audit_events"
-    read_timeout 0.1
-    basic_auth MnoEnterprise.tenant_id, MnoEnterprise.tenant_key
+    @@listeners = []
+    @@listeners << AuditEventsListener.new
+    @@listeners << IntercomEventsListener.new
```
ouranos: Why not  ?
```ruby
@@listeners = [
  AuditEventsListener.new,
  IntercomEventsListener.new
]
```

api/mno-enterprise-api.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'coffee-rails', '~> 4.1'
   s.add_runtime_dependency 'health_check', '~> 1.5'
   s.add_runtime_dependency 'httparty', '~> 0.13.7'
-
+  s.add_runtime_dependency 'intercom', '~> 3.5.4'
```
ouranos: We want an optional dependency on it (like sparkpost)

api/lib/mno_enterprise/intercom_events_listener.rb
```diff
@@ -0,0 +1,79 @@
...
+
+module MnoEnterprise
+
+  class IntercomEventsListener
+
+    attr_accessor :intercom
+
+    def initialize
+      if MnoEnterprise.intercom_api_key && MnoEnterprise.intercom_api_secret
+        self.intercom = ::Intercom::Client.new(app_id: MnoEnterprise.intercom_api_key, api_key: MnoEnterprise.intercom_api_secret)
```
ouranos: A bit of mixup between the app id and the api_key (probably because of way the tenant_id work?)

api/lib/mno_enterprise/intercom_events_listener.rb
```diff
@@ -0,0 +1,79 @@
...
+        case key
+          when 'user_update'
+            self.update_intercom_user(u)
+            return
+          when 'app_destroy'
+            event_name = 'deleted-app-' + object.app.nid
+          when 'app_add'
+            event_name = 'added-app-' + object.app.nid
+          else
+            event_name = key.gsub!(/_/, '-')
```
ouranos: I can show off a new trick I learned! I you don't use the regex capture groups in the replacement or if like here you're just doing a string substitution, `#tr` is order of magnitude faster than `#gsub`

See [this benchmark](https://github.com/JuanitoFatas/fast-ruby/blob/master/code/string/gsub-vs-tr.rb):

```
Warming up --------------------------------------
         String#gsub    49.395k i/100ms
           String#tr   134.293k i/100ms
Calculating -------------------------------------
         String#gsub    542.159k (±23.6%) i/s -      2.519M in   5.001670s
           String#tr      2.346M (± 5.8%) i/s -     11.683M in   5.001330s

Comparison:
           String#tr:  2345516.8 i/s
         String#gsub:   542158.9 i/s - 4.33x  slower
```

api/lib/mno_enterprise/intercom_events_listener.rb
```diff
@@ -0,0 +1,79 @@
...
+      else
+        self.intercom = nil
+      end
+
+    end
+
+    def info(key, current_user_id, description, metadata, object)
+      return unless self.intercom
+      u = User.find(current_user_id)
+      begin
```
ouranos: I don't think you need this `begin`, the `rescue Intercom::IntercomError` will apply to the whole method and it's easier to read.

ouranos:
@x4d3 see my changes.

I think we still miss the finishedSignUp event and the AddApp event doesn't have the same payload.
See: [intercom-svc.js.coffee](https://github.com/maestrano/Maestrano/blob/develop/app/assets/javascripts/angular/services/intercom-svc.js.coffee#L98)

api/lib/mno_enterprise/event_logger.rb
```diff
@@ -2,23 +2,14 @@
 
 module MnoEnterprise
   class EventLogger
-    include HTTParty
-    base_uri "#{MnoEnterprise.mno_api_private_host || MnoEnterprise.mno_api_host}/api/mnoe/v1/audit_events"
-    read_timeout 0.1
-    basic_auth MnoEnterprise.tenant_id, MnoEnterprise.tenant_key
+    @@listeners = [AuditEventsListener.new]
+    @@listeners << IntercomEventsListener.new if defined?(Intercom)
```
x4d3: nice!

x4d3:
@ouranos I've added specs.
I had to add intercom as a dependency in `mno-enterprise/api/mno-enterprise-api.gemspec` . Otherwise the `require 'intercom'` was failing in `mno-enterprise/api/lib/mno_enterprise/intercom_events_listener.rb`

PR #85, opened 2016-09-21
[MNO-271] Add support for Intercom secure mode

ouranos:
Intercom need a server side generated HMAC of the user_id or email when
secure mode is enabled.

See: https://docs.intercom.com/configure-intercom-for-your-product-or-site/staying-secure/enable-secure-mode-on-your-web-product


BrunoChauvet:
:+1:

PR #88, opened 2016-09-22
[MNOE-165] Does not permit to install a second app if the app is not multi_instantiable

x4d3:
See https://github.com/maestrano/maestrano-hub/pull/153

add multi_instantiable and subcategories to App model.

### DIFF
--- a/api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
+++ b/api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
@@ -8,6 +8,7 @@ json.is_star_ready app.star_ready?
 json.is_connec_ready app.connec_ready?
 json.is_coming_soon app.coming_soon?
 json.single_billing app.single_billing?
+json.multi_instantiable app.multi_instantiable
 
 if app.logo
   json.logo app.logo.to_s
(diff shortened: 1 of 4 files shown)


x4d3:
Done.

api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
```diff
@@ -8,6 +8,8 @@ json.is_star_ready app.star_ready?
 json.is_connec_ready app.connec_ready?
 json.is_coming_soon app.coming_soon?
 json.single_billing app.single_billing?
+json.multi_instantiable app.multi_instantiable
+json.subcategories app.subcategories
```
ouranos: Do we need the subcategories here?
If not it's better to keep the response smaller.

x4d3: yes we do. See: https://github.com/x4d3/mno-enterprise-angular/blob/e44298f3ae80534941c8c0a11a80dbf68fe57932/src/app/views/marketplace/marketplace-app.controller.coffee#L75

ouranos: Fair enough :+1:

ouranos:
To be merged once maestrano/maestrano-hub#153 is merged

PR #89, opened 2016-09-24
Fix Her monkey patch #first_or_create/initialize

ouranos:
Really simple patch to unwrap filter when creating/initializing.

Specs were a bit harder :)


BrunoChauvet:
:+1:

PR #90, opened 2016-09-24
[#73] Improve perf for dashboard#organizations

### LINKED ISSUE #73: Improve performance for `dashboard.organizations`
Hey there, 

I recently run into this code and I think it can create performance issue for dashboard request:
https://github.com/maestrano/mno-enterprise/blob/master/core/app/models/mno_enterprise/impac/dashboard.rb#L21

```
    def organizations
      self.organization_ids.map do |uid|
        MnoEnterprise::Organization.find_by(uid: uid)
      end
    end
```

Base on the code, it will call `MnoEnterprise::Organization.find_by(uid: uid)` for every `organization_id`, and since `MnoEnterprise::Organization.find_by` will call external services, it will slow down this method a lot when there are lots organization_ids.

I think we can change the `organizations` method to use `where` instead of `find_by`

ouranos:
- Remove N+1
- Fix indentation

This fixes #73

### DIFF
--- a/core/app/models/mno_enterprise/impac/dashboard.rb
+++ b/core/app/models/mno_enterprise/impac/dashboard.rb
@@ -3,31 +3,29 @@ module MnoEnterprise
 
     attributes :name, :widgets_order, :organization_ids, :widgets_templates, :currency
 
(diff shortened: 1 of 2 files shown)


BrunoChauvet:
Good catch :star:

PR #92, opened 2016-09-26
[MNOE-154] - Fixed bug in phone number

clemthenem:
Fixed a bug not displaying user's phone number on the sign up page.


ouranos:
All good. Split it in 3 PR with different target branches.
#96 #97 #98

PR #93, opened 2016-09-26
[MNOE-62] - Feature/terms page

clemthenem:
Updated the terms page to fit Maestrano terms.
Added some cache with @ouranos for the app list.


ouranos:
Closing to replace with #103

Only kept relevant commits, improved cache_key and squashed commits.

PR #102, opened 2016-09-30
Asynchronous Event Logging & improve doc

ouranos:
Using ActiveJob, so won't be asynchronous by default but the host app
can use any queue.


BrunoChauvet:
Looks OK given the condition on `Rails.env.test?` will be refactored with the specs at a later stage.

alachaum:
Looks good to me as well.

PR #83, opened 2016-09-16
WIP: Impac alert porting

xaun:
Specs still need to be completed..

@x4d3 could you please review? Thanks

@cesar-tonnoir @ouranos

PR #106, opened 2016-10-05
Impac alert porting

cesar-tonnoir:
Application of comments from @ouranos on https://github.com/maestrano/mno-enterprise/pull/83

#### Pending questions
* @xaun I may be wrong but I don't see the shared examples added in core/lib/mno_enterprise/testing_support/jpi_v1_test_helper.rb being used anywhere in the specs?


xaun:
@cesar-tonnoir you're correct, they are not being used. I guess I just left them in because they're handy methods. Feel free to remove.

PR #108, opened 2016-10-05
Improve perf in json views

alachaum:


api/app/views/mno_enterprise/jpi/v1/teams/_team.json.jbuilder
```diff
@@ -1,20 +1,26 @@
+org = @parent_organization || team.organization
+@all_apps ||= MnoEnterprise::App.all.to_a
```
ouranos: Should we filter with app_instances app nid?

ouranos:
@alachaum rebased from master and added a few other improvments.

@x4d3 could you review please?

PR #109, opened 2016-10-05
Intercom - Add more information when app is launched

x4d3:


ouranos:
Could you update the spec?

PR #110, opened 2016-10-05
[MNOE-138][MNOE-168] Fix double provisioning issue

ouranos:
Replaces #95 and should work instead of #105 too

@x4d3 FYI


BrunoChauvet:
:+1:

api/lib/mno_enterprise/concerns/controllers/provision_controller.rb
```diff
@@ -51,6 +51,9 @@ def new
   # POST /provision
   # TODO: check organization accessibility via ability
   def create
+    # Avoid double provisioning: previous url would be "/provision/new?apps[]=vtiger&organization_id=1"
+    session.delete('previous_url')
```
x4d3: Smart !

PR #118, opened 2016-10-07
[MNO-325] Sign out not always works

x4d3:
I don't know why the current_user method of mnoe is relying on the user being authenticated only for update and  :update_password methods


x4d3:
@ouranos Could we merge this ?

PR #119, opened 2016-10-07
[MNOE-177] UI: SignUp Page: Error message displayed when email is already registered

x4d3:
` validates_uniqueness_of email ` was useless. The User Model is using the devise option validatable which  performs the uniqueness validation

 http://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Validatable
 http://stackoverflow.com/questions/13897844/how-to-remove-duplicate-validation-errors


x4d3:
@ouranos Could you please review

Code Review

PR #123, opened 2016-10-11
[Request] Remove auto devise.skip_trackable

benpritchard:


ouranos:
Hi @benpritchard,

Could you provide more context on this one?

Thanks

benpritchard:
The problem as reported
-------------------------

Hi All,
I have opened fresh browser sessions to both UAT and Production this morning (after having the weekend off) and the site has recognised me as being logged in. It seems the session timeout of an hour is no longer in place? Can I please get some one to check this.

The investigation
-----------------

The timeout functionality comes from a devise module called Timeoutable.

It can be skipped per request by adding a request.env variable devise.skip_timeoutable, in which case none of the below is relevant.

When not totally skipped, the module adds a last_request_at var to the session and updates it each time a request is received unless:

there is a request.env var devise.skip_trackable == true, or
it has passed the timeout limit AND request.env var devise.skip_timeout != true (in which case it signs the user out)
As seen here Mno applies the devise.skip_trackable == true to all JSON GET requests, to avoid updating the timestamp, but any POST/PUT/DELETE should trigger the update since it will be false.

As seen here the Devise::SessionsController applies the devise.skip_timeout == true to the create action which also happens to be a POST.

Therefore if the first request after the session should have timed out is a a post to sessions#create, it will not force a logout (devise.skip_timeout == true) but will update the timestamp (devise.skip_trackable != true), thus extending the session.

This should not normally be a problem since the fact that the user is logging in (posting to sessions#create) should result in a valid session.

However the issue comes about because of the angular library angular-devise which is used as a wrapper for Devise. The front end uses this library to retrieve the current user based on the browser's session. In order to do this view code angular-devise sends a login request (a post to sessions#create), which in the above scenario will result in an extension of the session.

Scenarios where this should not happen:

any other request is received before the post to sessions#create
the browser has been closed (session cleared)
the browser is not reloaded (angular-devise should retain the User info and not need to make a login request)
Therefore if the browser has been left open (even if tab was closed) and a direct load or a refresh is performed and the first request happens to be the login request (likely) then the session will be extended.

To fix:

I suggest we add request.env["devise.skip_trackable"] = true to sessions#create to prevent it updating the timestamp.

ouranos:
Fixed in #578

PR #125, opened 2016-10-13
[MNOE-170] - Purchase Workflow

clemthenem:
Added an endpoint to create app_instances (see other pull request on MNOE angular)
Added a couple of css variables missing triggering errors.


api/app/controllers/mno_enterprise/jpi/v1/app_instances_controller.rb
```diff
@@ -7,7 +7,14 @@ def index
         i.active? && i.updated_at > Time.at(timestamp) && can?(:access,i)
       end
     end
-    
+
+    # POST /mnoe/jpi/v1/organization/1/app_instances
+    def create
+      authorize! :manage_app_instances, parent_organization
+      @app_instance = parent_organization.app_instances.create(product: params[:nid])
```
ouranos: No need for instance variable as you don't have a view.

PR #127, opened 2016-10-13
Fix deserialization error of deleted Object in EventLoggerJob

x4d3:
Delete object (for example widget) could not be found during deserialization in the EventLoggerJob and nil exception were happening.

### DIFF
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -261,4 +261,7 @@ module MnoEnterprise
       user
     end
   end
+
+
+
 end
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -261,7 +261,4 @@ module MnoEnterprise
       user
     end
   end
-
-
-
 end
--- a/api/app/controllers/mno_enterprise/jpi/v1/admin/invites_controller.rb
+++ b/api/app/controllers/mno_enterprise/jpi/v1/admin/invites_controller.rb
@@ -14,7 +14,7 @@ module MnoEnterprise::Jpi::V1::Admin
         send_org_invite(invite)
       end
 
-      MnoEnterprise::EventLogger.info('user_invite', current_user.id, 'User invited', {user_email: user.email, account_name: @organization.name}, user)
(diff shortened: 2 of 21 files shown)


x4d3:
@ouranos Could you please review

ouranos:
@x4d3 I still think there's a problem with the method calls:

5 args
```
MnoEnterprise::EventLogger.info('app_destroy', current_user.id, 'App destroyed', {name: app_instance.name, app_nid: app_instance.app.nid}, app_instance)
```

Expecting 6 args without default:
```
def info(key, current_user_id, description, subject_type, subject_id, metadata)
```

Could you also rebase from `master`?

x4d3:
@ouranos MnoEnterprise.EventLogger.info is taking 5 arguments
`def self.info(key, current_user_id, description, metadata, object)`

EventLoggerJob, IntercomEventsListener and AuditEventsListener are taking 6.

```
def self.info(key, current_user_id, description, metadata, object)
  formatted_metadata = format_metadata(metadata, object)
  subject_type = object.class.name
  subject_id = object.id
  # TODO: improve
  # Bypass Job queuing in specs or we'd have to stub lots of Her call for the deserialization
  if Rails.env.test?
    self.send_info(key, current_user_id, description, subject_type, subject_id, formatted_metadata)
  else
    MnoEnterprise::EventLoggerJob.perform_later('info', key, current_user_id, description, subject_type, subject_id, formatted_metadata)
  end
rescue ActiveJob::SerializationError
  Rails.logger.warn "[MnoEnterprise::EventLogger] Serialization error, skipping #{key} event"
end
```

ouranos:
Ok my bad. Looks good then.

Instead of passing a hash for the metadata, why not rely on the automatic call to `#to_audit_event` in `formatted_metadata`?

x4d3:
@ouranos Could you please review again, I took your idea of using 'to_audit_event' and I improved the code.

core/app/models/mno_enterprise/user.rb
```diff
@@ -261,4 +261,7 @@ def self.create_from_omniauth(auth, opts = {})
       user
     end
   end
+
+
+
```
ouranos: Empty lines :scream_cat:

PR #131, opened 2016-10-19
[MNOE-206] Expose free_trial_end_at to mnoe angular

hedudelgado:


hedudelgado:
@x4d3 @ouranos

api/app/views/mno_enterprise/jpi/v1/app_instances/_resource.json.jbuilder
```diff
@@ -8,6 +8,15 @@ json.oauth_keys_valid app_instance.oauth_keys_valid
 #json.microsoft_trial_url app_instance.microsoft_trial_url
 json.created_at app_instance.created_at
 
+if app_instance.per_user_licence?
+	json.per_user_licence app_instance.per_user_licence
+	json.licences_count app_instance.user_licences.active.count
```
ouranos: As mentioned via email, we only need the user licence count in mnoe.
So I'd suggest to do the count in MnoHub and expose is as a `active_licences_count` field on the `AppInstance` to avoid extra requests.

hedudelgado:
https://github.com/maestrano/maestrano-hub/pull/289

hedudelgado:
@ouranos

hedudelgado:
@ouranos I have moved the licence counting to the backend and exposing. You comment is gone because I squashed and old commit to removed something I should not have commited.

This was your comment.

+if app_instance.per_user_licence?
+	json.per_user_licence app_instance.per_user_licence
+	json.licences_count app_instance.user_licences.active.count
  @ouranos
ouranos 21 days ago Maestrano member
As mentioned via email, we only need the user licence count in mnoe.
So I'd suggest to do the count in MnoHub and expose is as a active_licences_count field on the AppInstance to avoid extra requests.

api/app/views/mno_enterprise/jpi/v1/app_instances/_resource.json.jbuilder
```diff
@@ -8,6 +8,15 @@ json.oauth_keys_valid app_instance.oauth_keys_valid
 #json.microsoft_trial_url app_instance.microsoft_trial_url
 json.created_at app_instance.created_at
 
+if app_instance.per_user_licence?
+	json.per_user_licence app_instance.per_user_licence
+	json.licences_count app_instance.active_licences_count
+end
+
+if app_instance.under_free_trial?
+  json.free_trial_end_at time_ago_in_words(app_instance.free_trial_end_at)
```
ouranos: @alexnoox I believe we should just send the date and do the "humanisation" in the frontend to be able to deal properly with i18n?

alexnoox: Yes, we use moment in the frontend. An API must return standard time.

hedudelgado: So I understand we would do
```
json.free_trial_end_at app_instance.free_trial_end_at
```
and then translate in the front end. Am I right?

@alexnoox  Do we have any example in our code using "moment" + "i18n"?

alexnoox: No, not from memory.

alexnoox: You can use the result of `moment()` in a locale parameter.

hedudelgado: **IN MNO-ENTERPRISE-ANGULAR**
Hey @alexnoox, once we have in the controller the result of moment(), lets say:
```
$scope.app.free_trial_end_at = moment(app.free_trial_end_at).format('dddd, MMMM Do YYYY, h:mm:ss a')
```
and in the view:
```
{{ 'mno_enterprise.templates.impac.dock.settings.free_trial' | translate }}: {{ app.free_trial_end_at }}
```
And displayed:
```
Free Trial: Thursday, April 6th 2017, 12:09:25 pm
```
how would you suggest to integrate i18n?

-I was thinking about using as result from moment something like 31-01-2017, and use the i18n the same way we did it with the pluralisation. (seems like a dirty approach)

-As other solution,  moment() seems to handle i18n:
https://momentjs.com/docs/#/i18n/

May you please hint me on this?
Thanks a lot.

alexnoox: Hi Eduardo,

I think you could you try a mix of angular-translate variable replacement (https://angular-translate.github.io/docs/#/guide/06_variable-replacement) and momentjs i18n to use the right format depending of the language used.

PR #135, opened 2016-10-24
Decorate controllers and removed unused ones

ouranos:


ouranos:
@hedudelgado FYI

api/lib/mno_enterprise/concerns/controllers/jpi/v1/app_instances_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  #==================================================================
+  # Instance methods
+  #==================================================================
+  # GET /mnoe/jpi/v1/organization/1/apps.json?timestamp=151452452345
+  def index
+    @app_instances = parent_organization.app_instances.select do |i|
+      i.active? && i.updated_at > Time.at(timestamp) && can?(:access,i)
+    end
+  end
```
BrunoChauvet: Could we improve this?

```ruby
@app_instances = parent_organization.app_instances.where(active: true).where("created_at >= ?", Time.at(timestamp)).select do |i|
  can?(:access,i)
end
```

ouranos: Good point. I'll see if I can merge #108 first

ouranos:
```ruby
@app_instances = parent_organization.app_instances.active.where("updated_at.gt" => Time.at(timestamp)).select do |i|
  can?(:access,i)
end
```

ouranos:
@BrunoChauvet rebased after perf improvements from #108

PR #138, opened 2016-10-25
Locales related to app provsionning fixes

alexnoox:
Better description when unauthorised to add an app


alexnoox:
@ouranos Related to https://github.com/maestrano/mno-enterprise-angular/pull/86

PR #146, opened 2016-11-07
[Perf] Add AssociationProxy#loaded?

ouranos:
Allow to check when association is already loaded to avoid unnecessary
queries.

@alachaum @BrunoChauvet should we extract this pattern 
```ruby
if self.organizations.loaded?
  self.organizations.to_a.find { |e| e.id == organization.id }
else
  self.organizations.where(id: organization.id).first
end
```
to a method on the HasManyAssociation? or is it too specific?


alachaum:
Not sure we should extract it. We should assume always assume that a call to the association method makes an API call. Will be easier to debug in case of issues.

PR #149, opened 2016-11-09
Fix built in email templates

ouranos:


ouranos:
Yes. This feature was introduced in 3.1 ;)

PR #154, opened 2016-11-21
[MNOE-243] Disable I18n by default in generator

ouranos:
Disable I18n when generating a new project as it won't be needed for
most projects.
This can still be enabled in the initializer if needed.


ouranos:
@adamaziz15 @hedudelgado FYI

hedudelgado:
👍

PR #156, opened 2016-11-24
[MNOE-241] Notification toastr after adding an app locales

alexnoox:


alexnoox:
@BrunoChauvet Please review

PR #158, opened 2016-11-28
[MNOE-250] Add conditional to disable protect_from_forgery

xaun:
Disabling the mno-enterprise AngularCSRF concern module is not enough for Impac! Workspace to fully work as `protect_from_forgery`([docs](http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html)) only allows GET requests.

@ouranos && @cesar-tonnoir please review.


xaun:
Hey guys, I don't think this change needs a spec right? I tried to spec it anyway with no success, show me the way :) :beers:

ouranos:
See comment in JIRA

PR #160, opened 2016-11-29
[MNOE-54][MNO-324] Manage admin role update backend

ouranos:
- Admin not allowed to edit Super Admin
- Last Super Admin cannot drop his role
- Spec the OrganizationController


ouranos:
@clemthenem FYI, refactored and added some specs.

PR #162, opened 2016-11-29
[MNOE-209] Unlimited failed attempts of login

x4d3:
user session is not returning either not found, either a user with password_not valid if the password is valid
Resource needed to be returned properly to be able to be updated in the devise-3.5.10/lib/devise/models/lockable.rb

### DIFF
--- a/core/lib/devise/strategies/remote_authenticatable.rb
+++ b/core/lib/devise/strategies/remote_authenticatable.rb
@@ -14,7 +14,7 @@ module Devise
 
 
         # mapping.to is a wrapper over the resource model
-        resource = mapping.to.authenticate(auth_params)
+        resource = mapping.to.new.remote_authentication(auth_params)
 
         return fail! unless resource
 
@@ -25,10 +25,9 @@ module Devise
         #
         # If the block returns true the resource will be loged in
         # If the block returns false the authentication will fail!
-        if  validate(resource){ resource.password_valid }
+        if validate(resource) { resource.password_valid }
           success!(resource)
         end
-
       end
     end
   end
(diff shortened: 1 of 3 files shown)


x4d3:
@ouranos Specs are passing

x4d3:
@ouranos I've rebased and squashed. Ready for final review.
You may test that the locking works on an express by adding in an express pointing to that version.
config/initializers/devise.rb
```Devise.setup do |config|
  config.lock_strategy = :failed_attempts
  config.maximum_attempts = 3
  config.last_attempt_warning = true
  config.paranoid = false
end
```

core/lib/devise/strategies/remote_authenticatable.rb
```diff
@@ -1,19 +1,20 @@
...
       # Method called by warden to authenticate a resource.
       def authenticate!
         # authentication_hash doesn't include the password
         auth_params = params[scope]
-        
+
+
         # mapping.to is a wrapper over the resource model
-        resource = mapping.to.new
+        resource = mapping.to.authenticate(auth_params)
```
ouranos: I'd keep `mapping.to.remote_authenticate(auth_params)`

core/lib/devise/strategies/remote_authenticatable.rb
```diff
@@ -24,9 +25,10 @@ def authenticate!
         #
         # If the block returns true the resource will be loged in
         # If the block returns false the authentication will fail!
-        if validate(resource){ resource = resource.remote_authentication(auth_params) }
+        if  validate(resource){ resource.password_valid }
```
ouranos: Extra space

core/lib/devise/strategies/remote_authenticatable.rb
```diff
@@ -24,9 +25,10 @@ def authenticate!
...
           success!(resource)
         end
+
```
ouranos: Extra line

x4d3:
@ouranos I've addressed your comments.

core/lib/devise/strategies/remote_authenticatable.rb
x4d3: you meant `mapping.to.new.remote_authentication(auth_params)` ?

PR #163, opened 2016-11-30
[IMPAC-408] Add pusher_key to MnoE impac config constant

xaun:
@ouranos @x4d3 please review

### DIFF
--- a/api/app/assets/javascripts/mno_enterprise/config.js.coffee.erb
+++ b/api/app/assets/javascripts/mno_enterprise/config.js.coffee.erb
@@ -1,9 +1,15 @@
 # Configuration module
 #
-# Recompile if the settings change
+# Dist: Recompile front-end if the settings change
+# Dev: Delete express project public/assets to ensure nothing is cached, bundle and restart.
 # = depend_on_config
 angular.module('mnoEnterprise.configuration', [])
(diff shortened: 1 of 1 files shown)


api/app/assets/javascripts/mno_enterprise/config.js.coffee.erb
```diff
@@ -1,9 +1,15 @@
...
-# Recompile if the settings change
+# Dist: Recompile front-end if the settings change
+# Dev: Delete express project public/assets to ensure nothing is cached, bundle and restart.
 # = depend_on_config
 angular.module('mnoEnterprise.configuration', [])
-  .constant('IMPAC_CONFIG', <%= Settings.impac.to_json %>)
+  .constant('IMPAC_CONFIG', {
+    settings: <%= Settings.impac.to_json %>,
+    keys: {
+      pusher_client: <%= ((k = ENV['pusher_key']) && k.to_json) || ''  %>
```
xaun: @ouranos maybe you can think of a better way I can do this? I want the value to be falsey should `env['pusher_key']` be nil.

xaun:
I have moved from accessing the env variables directly to using Settings, so alterations to mno-enterprise should no longer be needed given [this PR](https://github.com/maestrano/impac-express/pull/12) can work that way.

PR #166, opened 2016-12-05
[MNOE-259] Add New Relic config to generator

ouranos:
@hedudelgado & @x4d3 please review

### DIFF
--- a/core/lib/generators/mno_enterprise/install/templates/config/newrelic.yml
+++ b/core/lib/generators/mno_enterprise/install/templates/config/newrelic.yml
@@ -1,0 +1,46 @@
+#
+# This file configures the New Relic Agent.  New Relic monitors Ruby, Java,
+# .NET, PHP, Python and Node applications with deep visibility and low
+# overhead.  For more information, visit www.newrelic.com.
+#
+# For full documentation of agent configuration options, please refer to
+# https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration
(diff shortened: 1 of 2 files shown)


core/lib/generators/mno_enterprise/install/templates/config/newrelic.yml
```diff
@@ -0,0 +1,46 @@
...
+# This file configures the New Relic Agent.  New Relic monitors Ruby, Java,
+# .NET, PHP, Python and Node applications with deep visibility and low
+# overhead.  For more information, visit www.newrelic.com.
+#
+# For full documentation of agent configuration options, please refer to
+# https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration
+
+common: &default_settings
+  # Required license key associated with your New Relic account.
+  license_key: <%%= ENV['NEWRELIC_LICENSE_KEY'] %>
```
hedudelgado: <%%=

ouranos: This one is tricky :)

`<%%` will be rendered as a literal `<%`.

I want the resulting file to have `<%= ENV['NEWRELIC_LICENSE_KEY'] %>` and not to interpolate  to `ENV['NEWRELIC_LICENSE_KEY']` which would probably be `nil`

For the other `<%= Rails.application.class.parent_name %>`, I want it to be replaced by the app name when the file is generated.

PR #168, opened 2016-12-09
[MNO-287] developer secction

hedudelgado:
@ouranos 

I spotted a little bug on the code while testing it. This should fix it.


hedudelgado:
https://github.com/maestrano/maestrano-hub/pull/294

PR #171, opened 2016-12-09
[MNO-287] Change developer registration event description

hedudelgado:
@ouranos
I spotted a little bug on the code while testing it. This should fix it.

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/current_users_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/current_users_controller.rb
@@ -38,7 +38,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::CurrentUsersController
   def register_developer
     @user = current_user
     if @user.update(developer: true)
-      MnoEnterprise::EventLogger.info('register_developer', current_user.id, "User developer register", @user)
+      MnoEnterprise::EventLogger.info('register_developer', current_user.id, "Developer registration", @user)
       render :show
     else
       render json: @user.errors, status: :bad_request


hedudelgado:
@ouranos 
I had to squash the commits and a open a new and different PR to re-apply the commits that were reverted, otherwise they would not have been applied.

ouranos:
Hi @hedudelgado I don't think you need this. Check with @x4d3.
Also could you find a better description for this event. Thanks

x4d3:
confirm that the last argument of the function (metadata) is optional. The method will automatically use the to_audit_event of the user class.

hedudelgado:
@ouranos

ouranos:
@x4d3 see if the change is needed?
Also for the description, I'd put something like "User registered as developer" or "Developer registration" to keep it consistent with other events?

x4d3:
I agree it is important to log the fact that a User registered as a developer. And I agree with the description "Developer registration""

hedudelgado:
@ouranos

ouranos:
Hi @x4d3 and @hedudelgado 

As mentioned above, we changed the method signature in `master` to be
```ruby
self.info(key, current_user_id, description, object, metadata = {})
``` 
So the metadata is optional and `object.to_audit_event` is called if metadata is not provided.

Please fix this PR to only amend the event description or explain how this is fixing a bug

Thanks

ouranos:
Hi @x4d3 and @hedudelgado, could you look at the above? Thanks

hedudelgado:
His @x4d3 and @ouranos , unable to reproduce the bug that I apparently had, I cleaned the commits up and only include the changes on the description.

PR #177, opened 2016-12-22
Comments pagination

hedudelgado:
@ouranos  @alexnoox 
Have a look at the code if possible please. I will work the specs out tomorrow.


api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
```diff
@@ -30,6 +30,19 @@ def show
...
+  # GET /mnoe/jpi/v1/marketplace/:id/app_comments
+  def app_reviews
+    res = MnoEnterprise::AppReview.where(reviewable_id: params[:id])
+    res.limit(params[:limit]) if params[:limit]
+    res.skip(params[:offset]) if params[:offset]
+    res.order_by(params[:order_by]) if params[:order_by]
+    res.where(params[:where]) if params[:where]
+    @app_reviews = res.all.fetch
+    response.headers['X-Total-Count'] = @app_reviews.metadata[:pagination][:count]
+    return render json: "could not find Reviews for app #{params[:id]}", status: :not_found unless @app_reviews
```
ouranos: No need to return this, the view will render an empty array and the frontend will deal with it.

api/app/views/mno_enterprise/jpi/v1/marketplace/app_reviews.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.number_reviews @reviews_number
```
ouranos: No need for this.
@reviews_number is not defined.
We already send the total count in the headers

api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
```diff
@@ -30,6 +30,19 @@ def show
...
+    render 'app_reviews'
```
ouranos: No need for this, Rails will look for the view based on the action name.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
hedudelgado: I added the empty array management in the front end, I thought it could be useful to have it in the backend too.. I see it wasn't 😅

ouranos:
Looks better thanks.
It's missing the specs so I'll add them and merge.

hedudelgado:
Thanks a lot. I did work on the spec a bit too, but did not have time to make it work so I discard the changes to commit the code. May you please tag me in the specs to have a look at it once you add them?
Thank you very much.

ouranos:
@BrunoChauvet Sanity check please?

@hedudelgado @x4d3 I've rebased the branch and refactored the API. I'll send more details via email.

PR #183, opened 2017-01-06
[MNO-436] notify user of changes

clemthenem:
- Notify users when password is updated.
- Sends an email to the old email when email changes.
- Sends an email to the new email when email changes.

Added specs (omg...)

### DIFF
--- a/api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
+++ b/api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
@@ -23,7 +23,11 @@ module MnoEnterprise::Concerns::Mailers::SystemNotificationMailer
 
   # ==> Devise Email
   # Description:
-  #   Email asking users to confirm their email
+  #   New user: Email asking users to confirm their email
+  #     OR
+  #   Existing user: 
+  #    - Email asking users (on their new email) to confirm their email change
+  #    - Email notifying users (on their old email) of an email change
   #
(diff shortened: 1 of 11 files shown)


core/spec/assertions_helper.rb
```diff
@@ -0,0 +1,20 @@
+require 'active_support/test_case'
```
ouranos: No need for ActiveSupport::TestCase helper

api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
```diff
@@ -32,12 +32,21 @@ def default_sender
   #   :confirmation_link
   #
   def confirmation_instructions(record, token, opts={})
-    template = record.confirmed? && record.unconfirmed_email? ? 'reconfirmation-instructions' : 'confirmation-instructions'
+    update_email = record.confirmed? && record.unconfirmed_email?
```
ouranos: Could you add more comments here to explain what's going on?
Either on the method doc or inline

clemthenem:
Hey @ouranos! 
I did the changed requested.
Let me know if you need anything else from me on this one.
Cheers

api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
clemthenem: Just finished to do that on the method doc

PR #185, opened 2017-01-12
Added automatic locales generation to the new build

clemthenem:
Added automatic locales generation to the new build :)


ouranos:
Why are we using a tmp folder now?

Could you rebase on 3.2 and change the target branch of the PR?
`git rebase --onto 3.2 HEAD^` should do the trick

PR #186, opened 2017-01-12
Application Rating & Review

ouranos:
@alexnoox could you look at the last commit?

I've added a feature flag and updated the Admin Panel to use it.

Full sanity check of the PR would be good too :)


api/app/views/mno_enterprise/jpi/v1/marketplace/index.json.jbuilder
```diff
@@ -1,4 +1,2 @@
-json.cache! ['v1', 'marketplace'], expires_in: 20.minutes do
-  json.categories @categories
-  json.apps @apps, partial: 'app', as: :app
-end
+json.categories @categories
```
ouranos: We'll do some perf analysis to see the impact of removing the cache

PR #191, opened 2017-01-13
Invoke locales:generate in frontend:build task

xaun:
Hey @ouranos I was getting sick of running `mnoe:frontend:update` and then `mnoe:locales:generate` so I thought I would add this as I can see no reason to not.. unless I'm missing something of course. I tried to write a spec for this but then realised there a lot of stuff that needs to be done because of yarn and such.. So seeing as there currently isn't a spec for this file currently I decided not to dive into that 😂 :beers:


ouranos:
Hi @xaun , 

Thanks for that. That's what @clemthenem tried to do in #187 but I now see that he did in the wrong task.
I'll redo it and fix the specs.

ouranos:
Actually, I'll need to extract the build process to a class to make testing it simpler.
We only have specs for the locales generation at the moment.

I've done your 2 PR in one : #192

PR #199, opened 2017-01-24
[MNOE-282] App comparison config

c3gdlk:
…unning_instances_count into the API response


core/lib/generators/mno_enterprise/install/templates/config/settings.yml
```diff
@@ -24,3 +24,7 @@ developer:
 # Enable Reviews in the marketplace
 reviews:
   enabled: false
+
+marketplace:
+  comparison:
```
x4d3: I would prefer apps_comparison

c3gdlk: @x4d3 this name came from the issue description. Please confirm that I need to use the `apps_comparison`, it's 5 min to change

x4d3: ok I did not see it. Ignore my remark then.

ouranos: All good. I'll default it to `false` though.

ouranos: Done in #201

PR #216, opened 2017-02-01
Feature - rename widget_category to endpoint (FE interface)

cesar-tonnoir:
backward/forward compatible with mno-hub
**not** backward-compatible with impac-angular < v1.5.0

PR #222, opened 2017-02-06
Feature - Rework impac interfaces

cesar-tonnoir:
**Warning: requires deployment of impac-angular >= 1.5.0-rc3**
Backward-compatible with mnohub
changes approved here: https://github.com/maestrano/mno-enterprise/pull/216


cesar-tonnoir:
@ouranos following our discussion of last week, this is to be merged in **master** (and not 3.2).
I will rebuild mnoe-demo pointing to the correct version of impac-angular accordingly.
Can you confirm one last time we don't have any other front-end to update (any other front-end that would point to master instead of 3.2) ?

PR #228, opened 2017-02-10
Update default packages version

ouranos:
Running the install from the master branch should use edge packages.

### DIFF
--- a/frontend/lib/tasks/mnoe_frontend_tasks.rake
+++ b/frontend/lib/tasks/mnoe_frontend_tasks.rake
@@ -9,8 +9,8 @@ require 'rake/clean'
 namespace :mnoe do
   namespace :frontend do
     # Default version
-    MNOE_ANGULAR_VERSION = "1.0.x"
-    IMPAC_ANGULAR_VERSION = "1.4.x"
+    MNOE_ANGULAR_VERSION = "master"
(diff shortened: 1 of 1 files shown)


frontend/lib/tasks/mnoe_frontend_tasks.rake
```diff
@@ -9,8 +9,8 @@ require 'rake/clean'
 namespace :mnoe do
   namespace :frontend do
     # Default version
-    MNOE_ANGULAR_VERSION = "1.0.x"
-    IMPAC_ANGULAR_VERSION = "1.4.x"
+    MNOE_ANGULAR_VERSION = "master"
+    IMPAC_ANGULAR_VERSION = "1.5.0-rc3"
```
adamaziz15: @ouranos should this be "v1.5.0-rc3"? When I don't have the v in there I get an abiguous argument error from git

ouranos:
I copied the line from the demo: https://github.com/maestrano/mno-enterprise-demo/blob/develop/package.json#L5

Seems to be working for me. shouldn't have to put the `v` as `yarn` should find the tag corresponding to a specific version.

adamaziz15:
@ouranos @alexnoox Yes, my mistake. NPM gets confused without the v but yarn is able to find the correct tag :ok_hand: LGTM

PR #230, opened 2017-02-17
Admin Panel build

alexnoox:
@ouranos FYI, this is a work in progress.


alexnoox:
```
$ time rake mnoe:admin:install
real	1m12.411s
user	0m56.277s
sys	0m15.603s
```

post-45283-ace-ventura-dance-party-gif-im-j1ti

PR #182, opened 2017-01-05
[MNOE-155] - Staff invite workflow

clemthenem:
- Fixed bugs:
  * Staff did not received a confirmation email
  * Reloading the staff page was not working

- Improvments:
  * Redirect an "admin" user when he changes its details to "staff" to avoid accessing the staff page
  * Stop the last admin user to change its status or delete himself (frontend check)
  * Exit button logout staff without organizations and redirect to the dashboard otherwise.

PR #234, opened 2017-02-22
[MNOE-155] Staff invite workflow

ouranos:
Same as #182 but merging in `3.2` branch


ouranos:
Closing as this is backporting #182 in the `3.2` branch which is no longer supported

PR #235, opened 2017-02-23
Add add_on field for marketplace apps

hedudelgado:
@ouranos @x4d3


ouranos:
@hedudelgado @x4d3 where is this used? Thanks

hedudelgado:
@ouranos 
We added that in a demo that exposes if the app is add_on or not in the marketplace. We also did it while ago for the dock.

PR #195, opened 2017-01-19
[MNOE-291] Fix widget is not getting saved

hedudelgado:

PR #241, opened 2017-03-01
[MNOE-291] KPI & widget saving issues

xaun:
@ouranos this issue is pretty serious for Impac! as when you add widgets or KPIs to a Dashboard, the changes are not retrievable via mnoe until the dashboard object has reloaded.

@hedudelgado put forward a PR here: https://github.com/maestrano/mno-enterprise/pull/195, maybe you prefer this solution? 

@cesar-tonnoir @x4d3

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
@@ -119,17 +119,18 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::KpisController
   private
 
     def dashboard
-      @dashboard ||= MnoEnterprise::Impac::Dashboard.find(params.require(:dashboard_id))
+      @dashboard ||= current_user.dashboards.find(params.require(:dashboard_id))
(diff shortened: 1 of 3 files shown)


ouranos:
Could you also squash commits and add `[MNOE-291]` in it?

ouranos:
Actually this is gonna be an issue when the dashboard is assigned to an Organization instead of an User.
There was some code done in custom projects to handle polymorphic owner but that hasn't been merged back into the core.
The controllers (dashboard, widget, kpis) will need to be refactored.

As this is a pressing issue, I'll temporarily merge a variation of #195.
Do you have an easy way of reproducing this in dev so I can test before merging? I had trouble with this last time.

Thanks

xaun:
@ouranos can you link the temporary merge commit in this PR? 

Reproducing it is inconsistent, but try creating a new dashboard, and adding a KPI, then immediately refreshing the page. This way, the dashboard should not be in memory returning a "cached" dashboard without the recently added kpi / widget.

I removed the widget code because yes, it is not being used. It is for the attach kpi onto a widget capability which was parked, and not ported to mnoe.

ouranos:
@xaun see https://github.com/maestrano/mno-enterprise/pull/291

If all good we can close this one

PR #245, opened 2017-03-03
Update UPGRADING.md

ouranos:
Completed the upgrade doc.

Preview here: https://github.com/ouranos/mno-enterprise/blob/feature/update-doc/UPGRADING.md

@alexnoox @hedudelgado does that work for you guys?


hedudelgado:
👍  LGTM 👍

PR #249, opened 2017-03-07
[MNOE-375]  Add Payment restrictions

ouranos:
Refactor update_billing to use strong_params


ouranos:
See maestrano/mno-enterprise-angular#172

PR #251, opened 2017-03-08
Fix cache refresh when there's a marshalling error

ouranos:
@alachaum not super happy about this fix but it seems that Her is making the user a singleton in some cases.

@fgourichon that would fix the error at org creation

### DIFF
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -200,6 +200,9 @@ module MnoEnterprise
     def refresh_user_cache
       self.reload
       Rails.cache.write(['user', self.to_key], self)
+    # singleton can't be dumped / undefined method `marshal_dump' for nil
+    rescue TypeError, NoMethodError
+      expire_user_cache
     end
 
     # Used by omniauth providers to find or create users


ouranos:
@fgourichon Could you test that locally? I don't have the config for it

fgourichon:
@ouranos not working for SSO...
Error is : 
```
undefined method `marshal_dump' for nil:NilClass
 module_eval <<-RUBY, __FILE__, __LINE__ + 1 
 def #{name}(*args, &block) 
 #{target_name}.send(#{name.inspect}, *args, &block) 
 end 
 RUBY 
 end 
```

```
her (0.7.6) lib/her/model/associations/association_proxy.rb:11:in `marshal_dump'
/home/fgourichon/.rvm/gems/ruby-2.3.1/bundler/gems/mno-enterprise-5f1025998c98/core/lib/her_extension/model/associations/association_proxy.rb:33:in `method_missing'
activesupport (4.2.7.1) lib/active_support/cache.rb:667:in `dump'
 
```

ouranos:
@fgourichon Ok I've added NoMethodError too.

alachaum:
Is it worth adding a retry? (Does it happen randomly?

Otherwise happy with the fix. Not ideal but at least it does the job.

ouranos:
I don't think adding a retry will help with anything as we already reload the object before.
We can reproduce every time with specific workflow (impersonation or SSO from a specific vendor) but with different error for each workflow.
I have the feeling it has to do with the way Her manipulate the object or the associations but haven't been able to track it down.

PR #258, opened 2017-03-16
Add Admin Panel feature flags

ouranos:
See https://github.com/maestrano/mnoe-admin-panel/pull/7

### DIFF
--- a/core/lib/generators/mno_enterprise/install/templates/config/settings.yml
+++ b/core/lib/generators/mno_enterprise/install/templates/config/settings.yml
@@ -40,11 +40,24 @@ organization_management:
 # Allow user to edit their information and password
 user_management:
   enabled: true
-# Disable Impersonation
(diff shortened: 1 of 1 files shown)


core/lib/generators/mno_enterprise/install/templates/config/settings.yml
```diff
@@ -40,11 +40,24 @@ organization_management:
...
 # Disable user registration
 devise:
   registration:
     disabled: false
+
+# Admin Panel Config
+admin_panel:
+  # Disable Impersonation
+  impersonation:
+    disabled: false
```
fgourichon: Should we add a # TODO to refactor this one ?

ouranos: The whole settings file need to be refactored down the track

PR #261, opened 2017-03-22
[MNOE-399] Audit Log

ouranos:
- (Admin) Export Audit Log to CSV
- Link AuditEvent to Organization when possible
- Audit Log in dashboard

***Requires*** https://github.com/maestrano/maestrano-hub/pull/466

### DIFF
--- a/api/app/views/mno_enterprise/jpi/v1/audit_events/_audit_event.json.jbuilder
+++ b/api/app/views/mno_enterprise/jpi/v1/audit_events/_audit_event.json.jbuilder
@@ -1,0 +1,5 @@
+json.extract! audit_event, :id, :key, :user_id, :description, :created_at, :user
+
+if audit_event.formatted_details.present?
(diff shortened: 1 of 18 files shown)


api/app/views/mno_enterprise/jpi/v1/audit_events/_audit_event.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.extract! audit_event, :id, :key, :user_id, :description, :created_at, :user
```
fgourichon: No organization ?

ouranos: Not on this one as we are already within an organization. That'd be redundant.

ouranos:
@x4d3 could you have a look at the 2 last commits? Just after @BrunoChauvet review

PR #276, opened 2017-04-05
[MNOE-358] Fix dev mailer

ouranos:
The mail adapter is set in a class variable of `MailAdapter`. This doesn't
work well with class reloading in development.

Use a `config.to_prepare` block to fix this issue.


ouranos:
This causes the following error in dev when sending email after the classes have been reloaded:
```
NoMethodError (undefined method `deliver' for nil:NilClass ...
```

Easy to reproduce by sending an email (eg: forgot password) then changing a class (eg: `ApplicationController`) and sending an email again.

PR #280, opened 2017-04-25
[MNOE-249] Fix unconfirmed user can reset password

hedudelgado:
@ouranos


api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
```diff
@@ -64,11 +64,13 @@ def confirmation_instructions(record, token, opts={})
   #   :reset_password_link
   #
   def reset_password_instructions(record, token, opts={})
-    MnoEnterprise::MailClient.deliver('reset-password-instructions',
-      default_sender,
-      recipient(record),
-      user_vars(record).merge(reset_password_link: edit_user_password_url(reset_password_token: token))
-    )
+    if record.confirmed?
```
x4d3: I would also add a check on the controller level and eventually raise an error. What do you think @ouranos

ouranos:
Replaced with #581

PR #281, opened 2017-04-27
[MNOE-428] Migrate mno-enterprise to MnoHub API V2 

x4d3:


api/lib/mno_enterprise/concerns/controllers/jpi/v1/app_instances_controller.rb
```diff
@@ -13,31 +13,34 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::AppInstancesController
...
+    else
+      MnoEnterprise::EventLogger.info('app_add', current_user.id, 'App added', app_instance.first)
+      head :created
+    end
   end
 
   # DELETE /mnoe/jpi/v1/app_instances/1
   def destroy
-    app_instance = MnoEnterprise::AppInstance.find(params[:id])
+    app_instance = Mno::AppInstance.find(params[:id])[0]
```
ouranos: `find` is always returning an array? Unlike AR/Her which distinguish `find` from `find_all`?
In this case, I'd prefer `first` than `[0]`

Or should we use `find_one`?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
```diff
@@ -95,6 +95,7 @@ def dashboard_params
         whitelisted[:settings] = params[:dashboard][:metadata] || {}
       end
       .except(:metadata)
+      .merge(owner_type: "User", owner_id: current_user.id)
```
ouranos: This will cause backward compatibility issues with some customer projects where the owner is an Org.
They've extended the `dashboards` method.
I can share the code customisation with you as I believe it's quite flexible.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/widgets_controller.rb
```diff
@@ -17,44 +17,45 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::WidgetsController
...
 
   # DELETE /mnoe/jpi/v1/impac/widgets/:id
   #   -> DELETE /api/mnoe/v1/widgets/:id
   def destroy
-    if widget.destroy
-      MnoEnterprise::EventLogger.info('widget_delete', current_user.id, 'Widget Deletion', widget)
+    return render_not_found('widget') unless widget
+    MnoEnterprise::EventLogger.info('widget_delete', current_user.id, 'Widget Deletion', widget)
+    widget.destroy
+    if widget.errors.empty?
```
ouranos: Could we have the logger in this block? Or do you put it before to avoid issue with its being deleted?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
ouranos: Not to much of an issue as we'll tag a major release for this

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
```diff
@@ -176,21 +161,13 @@ def member
       @member ||= begin
         email = params.require(:member).require(:email)
         # Organizations are already loaded with all users
-        organization.users.to_a.find { |u| u.email == email } ||
-          organization.org_invites.active.where(user_email: email).first
+        organization.users.find { |u| u.email == email } ||
+        organization.orga_invites.find { |u| u.status == 'pending' && u.user_email = email}
```
ouranos: `u.user_email == email`

api/app/views/mno_enterprise/jpi/v1/organizations/_member.json.jbuilder
```diff
@@ -1,11 +1,11 @@
...
+if member.is_a?(Mno::User)
   json.id member.id
   json.entity 'User'
   json.name member.name
   json.surname member.surname
   json.email member.email
-  json.role member.role
-elsif member.is_a?(MnoEnterprise::OrgInvite)
+  json.role organization.role(member)
+elsif member.is_a?(Mno::OrgaInvite)
```
x4d3: yes. It is OrgaInvite in MnoHub. Let's have consistency :)

api/lib/mno_enterprise/concerns/controllers/jpi/v1/app_instances_controller.rb
```diff
@@ -13,31 +13,34 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::AppInstancesController
   #==================================================================
   # Instance methods
   #==================================================================
-  # GET /mnoe/jpi/v1/organization/1/apps.json?timestamp=151452452345
+  # GET /mnoe/jpi/v1/organization/1/apps.json?tg=151452452345
```
x4d3: ok. will do

api/lib/mno_enterprise/concerns/controllers/jpi/v1/app_instances_controller.rb
x4d3: yes, everything is a query for him, and therefore find return an array (I am not totally agains the approach, if you consider that everything is a collection, you can factorise code)

It is explained here: https://github.com/chingor13/json_api_client/issues/75

we should indeed use my lovely find_one

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/widgets_controller.rb
x4d3: You got it right. widget.id is empty after you call destroy. Quite annoying actually.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
x4d3: thanks !

api/mno-enterprise-api.gemspec
```diff
@@ -31,6 +31,8 @@ Gem::Specification.new do |s|
   # Lock sprocket version
   s.add_dependency 'sprockets-rails', '~> 2.3'
 
+  s.add_dependency 'json_api_client', '~> 1.3'
```
ouranos: Shouldn't that be a `mnoe-core` dependency?

api/spec/controllers/mno_enterprise/deletion_requests_controller_spec.rb
```diff
@@ -1,5 +1,10 @@
 require 'rails_helper'
 
+
+def main_app
```
ouranos: Is this needed?

core/app/models/mno/audit_event.rb
```diff
@@ -0,0 +1,26 @@
+module Mno
+  class AuditEvent < BaseResource
+    property :created_at, type: :time
+    property :updated_at, type: :time
+
+    def formatted_details
+      case details
+        when String
```
ouranos: Indent `when` as deep as `case` :policeman:

core/app/models/mno/base_resource.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  class CustomParser < ::JsonApiClient::Parsers::Parser
+    def self.parameters_from_resource(params)
+      hash = super
+      parse_types(hash)
+    end
+
+    def self.parse_types(res)
+      case res
+        when Array
```
ouranos: Indent `when` as deep as `case` :policeman:

core/app/models/mno/base_resource.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    def ==(o)
+      o.class == self.class && o.attributes == attributes
+    end
+
+  end
+end
+
+Mno::BaseResource.connection do |connection|
+  connection.use Faraday::Request::BasicAuthentication, MnoEnterprise.tenant_id, MnoEnterprise.tenant_key #
```
ouranos: Extra `#`

core/app/models/mno/identity.rb
```diff
@@ -0,0 +1,11 @@
+module Mno
+  class Identity < BaseResource
+    property :created_at, type: :time
+    property :updated_at, type: :time
+    property :user_id
+
+    def self.find_for_oauth(auth)
+        find_by_or_create(uid: auth.uid, provider: auth.provider)
```
ouranos: Indentation

core/app/models/mno/organization.rb
```diff
@@ -0,0 +1,77 @@
...
+      metadata && metadata['payment_restriction']
+    end
+
+    def role(user)
+      relation = self.orga_relation(user)
+      return relation.role if relation
+    end
+
+    def orga_relation(user)
+      self.orga_relations.find {|r|
```
ouranos: One liner or `do...end` :policeman:

core/app/models/mno/user.rb
```diff
@@ -0,0 +1,212 @@
...
+    property :locked_at, type: :time
+    property :last_sign_in_ip
+
+    define_model_callbacks :validation #required by Devise
+    define_model_callbacks :update #required by Devise
+    define_model_callbacks :create #required by Devise
+    define_model_callbacks :save #required by Devise
+
+
+    class RemoteUniquenessValidator < ::ActiveModel::EachValidator
```
ouranos: Not a big fan of nested class. Can you have that in a separate class like `core/lib/her_extension/validations/remote_uniqueness_validation.rb`?

core/app/models/mno/user.rb
```diff
@@ -0,0 +1,212 @@
...
+      super
+    end
+
+
+
+    #================================
+    # Validation
+    #================================
+    #
+    # if Devise.password_regex
```
ouranos: Validation?

core/lib/her_extension/middleware/mnoe_api_v1_parse_json.rb
```diff
@@ -16,23 +16,23 @@ def parse(body)
...
-          hash = res.dup
-          hash.each do |k,v|
-            hash[k] = parse_types(v)
-          end
-          return hash
-        when res.is_a?(String) && res =~ /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}/i
-          return Time.iso8601(res)
-        else
-          return res
+          when res.kind_of?(Array)
```
ouranos: Indent `when` as deep as `case` :policeman: 

Given that's Her related stuff, I'm sure it's Rubymine :)

core/lib/json_api_client_extension/json_api_client_orm_adapter.rb
```diff
@@ -0,0 +1,53 @@
...
+    def get!(id)
+      res = klass.includes(:deletion_requests, :organizations, :orga_relations, :dashboards).find(wrap_key(id)).first
+      raise JsonApiClient::Errors::ResourceNotFound, "resource not found" unless res
+      res
+    end
+
+    # @see OrmAdapter::Bax``
+    def get(id)
+      res = klass.includes(:deletion_requests, :organizations, :orga_relations, :dashboards).find(wrap_key(id))
+      if(res.errors && res.errors.first && res.errors.first.code != "404")
```
ouranos: See my ruby 2.3 comment, `&` ftw!

core/lib/mno_enterprise/core.rb
```diff
@@ -156,6 +158,13 @@ def host_url(path,opts = {})
   mattr_reader :mnoe_api_v1
   @@mnoe_api_v1 = nil
 
+
+  mattr_accessor :mno_api_v2_root_path
+  @@mno_api_v2_root_path = "/api/mnoe/v2"
+
+  mattr_reader :mnoe_api_v2
+  @@mnoe_api_v2 = nil
```
ouranos: What is this one used for?

core/mno-enterprise-core.gemspec
```diff
@@ -47,6 +47,8 @@ Gem::Specification.new do |s|
   # Config files per environment
   s.add_dependency 'config', '~> 1.0', '< 1.3'
 
+  s.add_dependency 'json_api_client', '~> 1.3'
```
ouranos: Add it next to her and faraday

api/app/views/mno_enterprise/jpi/v1/app_instances/_resource.json.jbuilder
```diff
@@ -4,17 +4,8 @@ json.stack app_instance.stack
 json.name app_instance.name
 json.status app_instance.status
 json.oauth_keys_valid app_instance.oauth_keys_valid
-#json.http_url app_instance.http_url
-#json.microsoft_trial_url app_instance.microsoft_trial_url
 json.created_at app_instance.created_at
 
-json.per_user_licence app_instance.per_user_licence
```
x4d3: Yes, it did not appear to be used in mno-enterprise-angular.

api/lib/mno_enterprise/concerns/controllers/deletion_requests_controller.rb
```diff
@@ -68,9 +68,9 @@ def freeze_account
         # Check that the deletion_request has the right status
         if @deletion_request.status == 'pending'
           @deletion_request.freeze_account!
-          format.html { redirect_to @deletion_request, notice: 'Your account has been frozen' }
+          format.html { redirect_to({action: :show, id: @deletion_request.id}, notice: 'Your account has been frozen') }
```
x4d3: No, but only because of the difference of namespace between the controller and the model.  This should be fixed if I put everything on the same namespace.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/app_instances_controller.rb
```diff
@@ -13,31 +13,33 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::AppInstancesController
...
   def create
     authorize! :manage_app_instances, parent_organization
-    app_instance = parent_organization.app_instances.create(product: params[:nid])
-    MnoEnterprise::EventLogger.info('app_add', current_user.id, 'App added', app_instance)
-    head :created
+    app_instance = parent_organization.provision_app_instance(params[:nid])
+    if app_instance.errors.any?
+      render json: app_instance.errors, status: :bad_request
+    else
+      MnoEnterprise::EventLogger.info('app_add', current_user.id, 'App added', app_instance.first)
```
x4d3: This is the way of working of json_api_client, everything returned is always an array.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/app_instances_controller.rb
```diff
@@ -13,31 +13,33 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::AppInstancesController
...
+    @app_instance = Mno::AppInstance.find_one(params[:id])
 
-    if app_instance
-      authorize! :manage_app_instances, app_instance.owner
-      MnoEnterprise::EventLogger.info('app_destroy', current_user.id, 'App destroyed', app_instance)
-      app_instance.terminate
+    if @app_instance
+      authorize! :manage_app_instances, @app_instance.owner
+      MnoEnterprise::EventLogger.info('app_destroy', current_user.id, 'App destroyed', @app_instance)
+      @app_instance = @app_instance.terminate.first
```
x4d3: I know, 

https://github.com/chingor13/json_api_client/issues/75

> This was intentional for a few reasons:
> 
> First, I always wanted to make a result set be available because there is extra meta data that comes back with the response. We could embed the meta data in a single resource object but that seems weird because it wouldn't be available on the resource objects fetched in a multi object response. I guess we could store a reference to the meta data but that also seems a bit janky.
> 
> Second, in the initial version of jsonapi, all responses returned arrays - even requests for a single resource. The rationale behind that was to make client side parsing easier - no matter what you were fetching, you could always iterate of the results like an array. There was a change (about a year ago?) to the spec that brought back the ability to respond with a single object or and array of objects as the main data. I wasn't a huge fan of it at the time.
> 
> Additionally, I really don't like the behavior in ActiveRecord where find can return a single result or an array depending on what input you give it. This departs from ActiveRecord but I'm ok with that here.
> 
> Hope this gives some insight into why it's doing this.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/alerts_controller.rb
```diff
@@ -12,20 +12,23 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::AlertsController
...
 
   # POST /jpi/v1/impac/kpis/:kpi_id/alerts
   def create
     return render_bad_request('attach alert to kpi', 'no alert specified') unless params.require(:alert)
-    return render_not_found('kpi') unless kpi_alert.kpi
+    return render_not_found('kpi') unless  Mno::Kpi.find_one(params.require(:kpi_id))
 
-    authorize! :manage_alert, kpi_alert
+    # TODO: Manage authorization
+    #authorize! :manage_alert, kpi_alert
```
x4d3: Yeah, I need to record that one. It's a bit difficult as the authorization was relying on the chain call from the widget to retrieve the user.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
```diff
@@ -143,7 +140,7 @@ def kpi_parent
 
     def kpi_create_params
       whitelist = [:dashboard_id, :widget_id, :endpoint, :source, :element_watched, {extra_watchables: []}]
-      extract_params(whitelist)
+      extract_params(whitelist).merge(dashboard_id: params[:dashboard_id])
```
x4d3: yeah, it should not be there. The whitelist parameters are taking it from kpi inside the prams. dashboard_id is on the top level. (see `extract_params` method)

api/lib/mno_enterprise/concerns/controllers/jpi/v1/teams_controller.rb
```diff
@@ -16,37 +16,29 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::TeamsController
...
-    # # Update permissions
-    if params[:team] && params[:team][:app_instances]
-      list = params[:team][:app_instances].select { |e| e != {} }
-      @team.set_access_to(list)
-    end
-
+    @team = Mno::Team.find_one(params[:id], :organization)
+    @parent_organization = Mno::Organization.find_one(@team.organization.id, :orga_relations)
+    authorize! :manage_teams, @parent_organization
+    @team.update_attributes(update_params)
```
x4d3: Yes it should.

api/lib/mno_enterprise/concerns/controllers/pages_controller.rb
```diff
@@ -23,15 +23,15 @@ module MnoEnterprise::Concerns::Controllers::PagesController
...
+    app_instance = Mno::AppInstance.where(uid: params[:id]).first
     MnoEnterprise::EventLogger.info('app_launch', current_user.id, 'App launched', app_instance)
     redirect_to MnoEnterprise.router.launch_url(params[:id], {wtk: MnoEnterprise.jwt(user_id: current_user.uid)}.reverse_merge(request.query_parameters))
   end
 
   # GET /loading/:id
   # Loading lounge - wait for an app to be online
   def loading
-    @app_instance = MnoEnterprise::AppInstance.where(uid: params[:id]).reload.first
+    @app_instance = Mno::AppInstance.find(uid: params[:id]).first
```
x4d3: `find` should not be used. It is either `where().first` when there is conditions or `find_one` to find by id.

api/spec/controllers/mno_enterprise/deletion_requests_controller_spec.rb
x4d3: The spec was not working in Rubymine without it.

core/lib/mno_enterprise/core.rb
x4d3: Nothing, good spot.

core/app/models/mno_enterprise/app.rb
```diff
@@ -1,69 +1,17 @@
...
-#  popup_description        :text
-#  stack                    :string(255)
-#  terms_url                :string(255)
-#  tags                     :text
-#  rank                     :integer
-#
-
 module MnoEnterprise
   class App < BaseResource
-    scope :active, -> { where(active: true) }
```
ouranos: We can no longer have `scope`? That was pretty handy.

core/app/models/mno_enterprise/audit_event.rb
```diff
@@ -1,5 +1,8 @@
 module MnoEnterprise
   class AuditEvent < BaseResource
+    property :created_at, type: :time
```
ouranos: Shouldn't this be on all models?

core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -1,96 +1,88 @@
...
-    # Class query methods
-    class << self
-      # Delegate the following methods to `scoped`
-      # Clear relation params for each class level query
-      [:all, :where, :create, :find, :first_or_create, :first_or_initialize, :limit, :skip, :order_by, :sort_by, :order, :sort].each do |method|
-        class_eval <<-RUBY, __FILE__, __LINE__ + 1
-          def #{method}(*params)
-            Her::Model::Relation.new(self).send(#{method.to_sym.inspect}, *params)
+
+  class CustomParser < ::JsonApiClient::Parsers::Parser
```
ouranos: Extract to separate file?

core/app/models/mno_enterprise/kpi.rb
```diff
@@ -0,0 +1,6 @@
+module MnoEnterprise
+  class Kpi < BaseResource
```
ouranos: No longer namespaced under `Impac`?

core/app/models/mno_enterprise/orga_invite.rb
```diff
@@ -0,0 +1,37 @@
...
+    property :updated_at, type: :time
+
+    custom_endpoint :accept, on: :member, request_method: :patch
+    custom_endpoint :decline, on: :member, request_method: :patch
+
+    def to_audit_event
+      self.attributes.slice(:team_id, :user_role, :user_email, :user_id, :referrer_id, :organization_id)
+    end
+
+    # TODO: specs
```
ouranos: Do we have specs for that in v2?

core/app/models/mno_enterprise/user.rb
```diff
@@ -1,59 +1,43 @@
...
-#  geo_city                       :string(255)
-#  website                        :string(255)
-#  api_key                        :string(255)
-#  api_secret                     :string(255)
-#
-
 module MnoEnterprise
   class User < BaseResource
-    include MnoEnterprise::Concerns::Models::IntercomUser if MnoEnterprise.intercom_enabled?
+    include MnoEnterprise::Concerns::Models::IntercomUser
```
ouranos: No condition anymore?

core/app/models/mno_enterprise/user.rb
```diff
@@ -71,143 +55,85 @@ class User < BaseResource
...
     # Return nil in case of failure
-    def self.authenticate(auth_hash)
-      u = self.post("user_sessions", auth_hash)
-
-      if u && u.id
-        u.clear_attribute_changes!
-        return u
+    def self.authenticate_user(auth_hash)
+      result = self.authenticate({data: {attributes: auth_hash}})
+      if result
```
ouranos: If we don't need the `clear_attribute_changes`:
```ruby
result = self.authenticate({data: {attributes: auth_hash}})
if (u = result&.first) && u.id
  u
end
```
no need for the nil at the end

core/lib/devise/models/remote_authenticatable.rb
```diff
@@ -12,7 +12,7 @@ module RemoteAuthenticatable
       # If the authentication fails you should return false
       #
       def remote_authentication(authentication_hash)
-        self.class.authenticate(authentication_hash) # call MnoEnterprise::User.authenticate
+        self.class.authenticate_user(authentication_hash) # call MnoEnterprise::User.authenticate
```
ouranos: The comment is not really needed

core/lib/devise/models/remote_authenticatable.rb
```diff
@@ -28,9 +28,11 @@ def send_password_change_notification
       # Overriden methods from Devise::Models::Authenticatable
       ####################################
       module ClassMethods
+
         # Flag to enable password change notification
         Devise::Models.config(self, :send_password_change_notification)
 
+
```
ouranos: Extra blank line

core/lib/json_api_client_extension/json_api_client_orm_adapter.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    # @see OrmAdapter::Base#get!
+    def get!(id)
+      res = klass.includes(:deletion_requests, :organizations, :orga_relations, :dashboards).find(wrap_key(id)).first
+      raise JsonApiClient::Errors::ResourceNotFound, "resource not found" unless res
+      res
+    end
+
+    # @see OrmAdapter::Bax``
```
ouranos: `Bax`?

core/lib/json_api_client_extension/json_api_client_orm_adapter.rb
```diff
@@ -0,0 +1,53 @@
...
+    def get(id)
+      res = klass.includes(:deletion_requests, :organizations, :orga_relations, :dashboards).find(wrap_key(id))
+      error = res&.errors&.first
+      if(error && error.code != '404')
+        raise  error.detail
```
ouranos: double space

core/lib/json_api_client_extension/validations/remote_uniqueness_validation.rb
```diff
@@ -0,0 +1,18 @@
+module JsonApiClientExtension
+  module Validations
+    class RemoteUniquenessValidator < ::ActiveModel::EachValidator
+      def validate_each(record, attribute, value)
+        list = record.class.where({attribute => value}).paginate(page: 1, per_page: 1).to_a
```
ouranos: can you do `record.class.find_first({attribute => value}).to_a` ?

and `list.any? { |e| e && e.id != record.id  }`

core/lib/mno_enterprise/concerns/models/app_instance.rb
```diff
@@ -76,38 +35,14 @@ module ClassMethods
...
-  # Send a request to terminate the AppInstance
-  # Alias of destroy
-  # TODO: specs
-  def terminate
-    self.destroy
-  end
-
-  # Return true if the instance can be considered active
-  # Route53 DNS propagation may take up to a minute, so we force a minimum of 60 seconds before considering the application online
-  def active?
```
ouranos: Shouldn't we keep all those instances methods (this and the one below)

core/lib/mno_enterprise/concerns/models/team.rb
```diff
@@ -42,26 +23,4 @@ module ClassMethods
   #==================================================================
   # Instance methods
   #==================================================================
-  # Add a user to the team
-  # TODO: specs
-  def add_user(user)
```
ouranos: This is no longer needed?

core/lib/mno_enterprise/core.rb
```diff
@@ -22,8 +22,10 @@
 require "her_extension/middleware/mnoe_raise_error"
```
ouranos: Let's get rid of this!

core/mno-enterprise-core.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency "her", "~> 0.7.3"
```
ouranos: Let's remove it!

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
x4d3: we could change the api to accept an owner_type externally, that could be User or Organization

api/app/views/mno_enterprise/jpi/v1/organizations/_current_user.json.jbuilder
```diff
@@ -2,4 +2,4 @@ json.id user.id
 json.name user.name
 json.surname user.surname
 json.email user.email
-json.role organization.members.to_a.find { |e| e.id == user.id }.role
+json.role user.role(organization)
```
x4d3: There is no performance issue, as the user is already loaded with it's organization.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
```diff
@@ -56,14 +56,14 @@ def create
...
     end
     # TODO: nest alert in as a param, with the current user as a recipient.
-    @kpi = kpi_parent.kpis.create(kpi_create_params)
-    unless kpi.errors?
+    @kpi = MnoEnterprise::Kpi.create(kpi_create_params)
+    if @kpi.errors.empty?
       # Creates a default alert for kpis created with targets defined.
       if kpi.targets.present?
-        current_user.alerts.create({service: 'inapp', impac_kpi_id: kpi.id})
+        MnoEnterprise::Alert.create({service: 'inapp', kpi_id: kpi.id, recipient_ids: [current_user.id]})
```
x4d3: It is in the recipient ids indeed, it was done this way on mnohub v1. I've simplified.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
```diff
@@ -31,14 +30,14 @@ def index
...
+    @app = MnoEnterprise::App.find_one(params[:id])
   end
 
   def app_relation
     if MnoEnterprise.marketplace_listing
-      MnoEnterprise::App.where('nid.in' => MnoEnterprise.marketplace_listing)
+      MnoEnterprise::App.where(nid: MnoEnterprise.marketplace_listing)
     else
-      MnoEnterprise::App.all
+      MnoEnterprise::App.where
```
x4d3: For some reasons the .all was not returning a query builder. The `where` is.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
```diff
@@ -135,26 +117,28 @@ def update_member
...
-    if organization.role == 'Admin'
+    if current_user.role(organization) == 'Admin'
       # Admin cannot assign Super Admin role
       raise CanCan::AccessDenied if attributes[:role] == 'Super Admin'
 
       # Admin cannot edit Super Admin
-      raise CanCan::AccessDenied if (member.is_a?(MnoEnterprise::User) && member.role == 'Super Admin') ||
-        (member.is_a?(MnoEnterprise::OrgInvite) && member.user_role == 'Super Admin')
-    elsif member.id == current_user.id && attributes[:role] != 'Super Admin' && organization.users.count {|u| u.role == 'Super Admin'} <= 1
+      raise CanCan::AccessDenied if (member.is_a?(MnoEnterprise::User) && organization.role(member) == 'Super Admin') || (member.is_a?(MnoEnterprise::OrgaInvite) && member.user_role == 'Super Admin')
```
x4d3: It never was, I will make it better.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/teams_controller.rb
```diff
@@ -74,19 +66,35 @@ def destroy
...
+    @parent_organization = MnoEnterprise::Organization.find_one(@team.organization.id, :orga_relations)
     render 'show'
   end
 
-  def team_params
-    params.require(:team).permit(:name)
+  def update_params
+    update = params.require(:team).permit(:name)
+    if params[:team] && params[:team][:app_instances]
+      list = params[:team][:app_instances].select { |e| e != {} }.map { |e| e['id'] }
```
x4d3: Thanks !

api/lib/mno_enterprise/concerns/controllers/org_invites_controller.rb
```diff
@@ -25,17 +25,17 @@ module ClassMethods
...
-
-    if @org_invite && !@org_invite.expired? && @org_invite.accept!(current_user)
-      redirect_path = add_param_to_fragment(redirect_path.to_s, 'dhbRefId', @org_invite.organization.id)
-      message = { notice: "You are now part of #{@org_invite.organization.name}" }
-      yield(:success, @org_invite) if block_given?
-    elsif @org_invite && @org_invite.expired?
-      message = { alert: "It looks like this invite has expired. Please ask your company administrator to resend the invite." }
-    else
-      message = { alert: "Unfortunately, this invite does not seem to be valid." }
+    message = { alert: 'Unfortunately, this invite does not seem to be valid.'}
```
x4d3: I'll add the TODO ;)

api/spec/controllers/mno_enterprise/impersonate_controller_spec.rb
```diff
@@ -9,18 +9,21 @@ module MnoEnterprise
...
-      api_stub_for(get: "/users/#{user2.id}", response: from_api(user2))
-      api_stub_for(put: "/users/#{user2.id}", response: from_api(user2))
+      stub_api_v2(:get, "/users/#{user.id}", user, %i(deletion_requests organizations orga_relations dashboards))
+      stub_api_v2(:get, "/users/#{user2.id}", user2, %i(deletion_requests organizations orga_relations dashboards))
+
+      stub_api_v2(:patch, "/users/#{user.id}")
+      stub_api_v2(:patch, "/users/#{user2.id}")
+
     end
+    before { stub_audit_events }
```
x4d3: I don't like the idea to stub it globally as it force the specs to be awayre that the audit_events is being stubbed. 
Also we could improve this stub to test that the right fields and comments are being sent.

api/spec/controllers/mno_enterprise/jpi/v1/app_instances_controller_spec.rb
```diff
@@ -6,85 +6,84 @@ module MnoEnterprise
...
       subject { post :create, organization_id: organization.id, nid: 'my-app' }
 
       before do
-        api_stub_for(post: "/organizations/#{organization.id}/app_instances", response: from_api(app_instance))
-        api_stub_for(get: "/organizations/#{organization.id}/app_instances", response: from_api([app_instance]))
+        stub_api_v2(:post, '/app_instances/provision', app_instance)
       end
-      it_behaves_like "jpi v1 protected action"
+      it_behaves_like 'jpi v1 protected action'
+      it {
```
x4d3: Sorry, Rubymine is displaying that with a warning, and I automatically replace the " by `

api/spec/controllers/mno_enterprise/jpi/v1/impac/kpis_controller_spec.rb
```diff
@@ -79,20 +77,21 @@ module MnoEnterprise
 
         it "creates the kpi" do
           subject
-          expect(assigns(:kpi)).to eq(kpi)
+          # TODO: check that the rendered kpi is the created one
+          # expect(assigns(:kpi)).to eq(kpi)
```
x4d3: Yes, you're right.

api/spec/spec_helper.rb
```diff
@@ -15,6 +15,8 @@
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 require "fakeweb"
+# Library for stubbing and setting expectations on HTTP requests in Ruby.
+require 'webmock/rspec'
```
x4d3: Kind of odd, but without fakeweb, he api was trying to do the real calls.
With it, there was an error explicitly describing which api was not stubbed, which was very handy.

core/app/models/mno_enterprise/app.rb
x4d3: I agree. One of the missing feature of https://github.com/chingor13/json_api_client

core/app/models/mno_enterprise/base_resource.rb
x4d3: :sob: :sob: :sob: 

yes...

core/app/models/mno_enterprise/kpi.rb
x4d3: No indeed, I don't see any advantage of the namespacing compared to the pain and suffering it is bringing, by usually not working, because it gets off the beaten track.

core/app/models/mno_enterprise/user.rb
x4d3: It could stay, but I had a lot of trouble making specs works because of it (don't really remember why)
And I thought that could be simplified. That is indeed debatable.

core/lib/json_api_client_extension/validations/remote_uniqueness_validation.rb
x4d3: I am not sure `find_first` exists. I will have a look.

core/lib/mno_enterprise/concerns/models/team.rb
x4d3: Needs to check, good spot.

x4d3:
Some of the TODO were already here, (for example the Ability on the Impac model)
> not a big fan of the let!(stub) is there a particular reason?
To be able to replace the stub for some particular case.
> remove her dependency (and all helpers files)?
Yep, I should do that in a separate commit indeed.
AppReview/AppComment => Needed for Opal project
> Yes, I had issue with polymorphism.

question:
> TSchema information in model is quite useful?
True, however I am not a big fan as it is not always up to date. I think I would prefer to use the attributes as a way to document the fields and  their types. What do you think ?
attributes to property is not 1 => 1 (eg: a big list is replaced by created_at/updated_at)
Yes, because they are not needed explicitly to work. I will add them (Maybe using MnoHub)
> the assocations are no longer explicit?
Unfortunately no, it is documented in MnoHub, but not visible in MnoE,

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
x4d3: but  the current code was using UserDashboard, and therefore was already creating a dashboard for which the owner was a user.

There is no backward issue.

api/app/views/mno_enterprise/jpi/v1/audit_events/index.json.jbuilder
```diff
@@ -1,4 +1,6 @@
 json.audit_events do
   json.array! @audit_events, partial: 'audit_event', as: :audit_event
 end
-json.metadata @audit_events.metadata
+# TODO: Port previous pagination metadata information ?
+# json.metadata @audit_events.metadata
```
x4d3: Ok. Removing it.

api/lib/mno_enterprise/concerns/controllers/deletion_requests_controller.rb
x4d3: Corrected.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
```diff
@@ -34,9 +34,10 @@ def create
     # TODO: enable authorization
```
x4d3: Could it be in another PR ?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
x4d3: I confirm,  there is ` alias all to_a` in the builder, and here I need a query builder.

api/spec/controllers/mno_enterprise/auth/confirmation_controller_spec.rb
```diff
@@ -33,17 +33,12 @@ module MnoEnterprise
 
       context 'confirmed user' do
         let(:user) { confirmed_user }
-
         context 'with a new email' do
           let(:email) { 'unconfirmed@example.com' }
-
-          before do
+          let!(:api_stubs) do
```
x4d3: I had issue with the stub beeing called twice, also the problem is that when stub_api_v2 is being called, it is the current state of the input that will be saved as json and stored in the stub. Therefore, is the input is changed in another block or let(), the stub won't be updated. It's not dynamic as the argument is saved serialized in the stub.

core/app/models/mno_enterprise/audit_event.rb
x4d3: Yes. I did not find a way to declare it automatically on all the class. Adding it to base_resource does not work, as it only populate the attributes of base_resource, not the child class.

core/app/models/mno_enterprise/orga_invite.rb
x4d3: Not yet. TODO

core/lib/devise/models/remote_authenticatable.rb
x4d3: Ok.

core/lib/json_api_client_extension/validations/remote_uniqueness_validation.rb
x4d3: find_first does not exist in json api client. :(

core/lib/mno_enterprise/concerns/models/app_instance.rb
x4d3: If they are not called. (and they are not), why should we ?

core/lib/mno_enterprise/concerns/models/team.rb
x4d3: It is used in organization_controller.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
```diff
@@ -120,10 +120,16 @@ def update_member
...
-
+      member_current_role = if member.is_a?(MnoEnterprise::User)
+                              organization.role(member)
+                            elsif member.is_a?(MnoEnterprise::OrgaInvite)
+                              member.user_role
+                            end
       # Admin cannot edit Super Admin
-      raise CanCan::AccessDenied if (member.is_a?(MnoEnterprise::User) && organization.role(member) == 'Super Admin') || (member.is_a?(MnoEnterprise::OrgaInvite) && member.user_role == 'Super Admin')
-    elsif member.id == current_user.id && attributes[:role] != 'Super Admin' && organization.orga_relations.count {|u| u.role == 'Super Admin'} <= 1
+      if member_current_role == 'Super Admin'
```
ouranos: This can now be done in one line with `if` as a modifier instead of a statement

api/spec/controllers/mno_enterprise/deletion_requests_controller_spec.rb
```diff
@@ -58,14 +58,13 @@ def main_app
 
       context 'when the request is pending' do
         it 'freezes the account' do
-          expect(controller.current_user).to receive(:current_deletion_request).and_return(deletion_req)
-          expect(deletion_req).to receive(:freeze_account!)
+          expect_any_instance_of(MnoEnterprise::DeletionRequest).to receive(:freeze_account!)
```
ouranos: I think rubocop prefers the previous code :)
Because you're testing you're freezing the correct deletion request

core/lib/mno_enterprise/concerns/models/app_instance.rb
ouranos: https://github.com/maestrano/mno-enterprise/blob/master/api/app/views/mno_enterprise/jpi/v1/app_instances/_resource.json.jbuilder#L11

api/spec/controllers/mno_enterprise/deletion_requests_controller_spec.rb
x4d3: but it won't be the same object, as the one passed is a one created from the parsing of the api result...

api/spec/spec_helper.rb
ouranos: If you're using webmock just for this purpose: `FakeWeb.allow_net_connect = false`
We don't need both.
Btw, I've switched all test domains to non routable TLD

api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
```diff
@@ -16,11 +16,17 @@ json.average_rating app.average_rating
 json.add_on app.add_on?
 json.running_instances_count app.running_instances_count
 
-json.shared_entities do
-  json.array! app.shared_entities do |shared_entity|
-    json.extract! shared_entity, :shared_entity_nid, :shared_entity_name, :write, :read
+if app.app_shared_entities.any?
+  json.app_shared_entities do
```
ouranos: The response is changing from `shared_entities` to `app_shared_entities`?

api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
```diff
@@ -16,11 +16,17 @@ json.average_rating app.average_rating
...
+    json.array! app.app_shared_entities do |shared_entity|
+      json.shared_entity_nid shared_entity.shared_entity.nid
```
ouranos: Why not use `json.extract!`? It's a bit more  concise?

api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
x4d3: The app contains a list of app_shared_entities in the v2 api.

api/app/views/mno_enterprise/jpi/v1/marketplace/_app.json.jbuilder
x4d3: because it's not taking the fields name by name, it is mapping 
shared_entity_nid to shared_entity.nid.

x4d3:
because in the MnoHub model, an app contains AppSharedEntities, that contains SharedEntity

PR #290, opened 2017-05-09
Add URL_CONFIG

fgourichon:
@ouranos for review


ouranos:
Actually, should we reuse the old router config to populate the settings.js? So that we keep the old behavior?

fgourichon:
Yes. I wasn't sure if the "config.router.after_sign_out_url" was going to
be available in the yml file. Can I use it straight away?

Francois Gourichon

*VP Engineering**Maestrano*

*Australia *I* USA *I* Dubai *I* Singapore* I* London*

Suite 504, 46 Market Street, Sydney, NSW 2000, Australia

francois.gourichon@maestrano.com  | www.maestrano.com
<http://maestrano.com/>

PR #291, opened 2017-05-10
[MNOE-291] Fix association cache bug

ouranos:
@cesar-tonnoir @xaun before I merge, could you point to my branch and confirm that this fixes the problem?
The cache will be slightly less efficient but we won't have stale data anymore.


ouranos:
This will supersede #195 #241

cesar-tonnoir:
Tested on my local - wasn't able to reproduce the bug. All good!

PR #295, opened 2017-05-15
Fix locale

hedudelgado:
@alexnoox


alexnoox:
@hedudelgado Thanks, good catch ;)
@ouranos FYI

PR #298, opened 2017-05-18
[MNO-643] Fix logos no readable on white background in invoices

hedudelgado:


hedudelgado:
@ouranos

core/app/pdf/mno_enterprise/invoice_pdf.rb
```diff
@@ -116,29 +117,18 @@ def metadata
...
 
-
     # Add a repeated header to the document
     def add_page_header
       title = Settings.payment.disabled ? 'Account Statement - ' : 'Monthly Invoice - '
       @pdf.repeat :all do
         @pdf.bounding_box([0, @pdf.bounds.top+@format[:header_size]], width: 540, height: @format[:footer_size]) do
           @pdf.float do
-            @pdf.image image_path('mno_enterprise/main-logo.png'), scale: 0.5
+            @pdf.image select_image, scale: 0.5
```
ouranos: The method name should be more explicit than `select_image`

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -0,0 +1,22 @@
+module MnoEnterprise
+  module ImageHelper
+    # Helper method to easily access and select the images
+    def select_image
```
ouranos: The method name is pretty obscure.

I think the helper could be called something like `ImageHelper#main_logo_white_bg`?
That'd be a lot clearer.

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+      main_logo_whitebg = "#{path}mno_enterprise/main-logo-whitebg.png"
+      return image(main_logo_whitebg) if image(main_logo_whitebg)
+
+      main_logo = "#{path}mno_enterprise/main-logo.png"
+      return image(main_logo) if image(main_logo)
+
+      return "#{MnoEnterprise::Engine.root}#{main_logo}"
+    end
+
+    def image(logo)
```
ouranos: Same here. No comments and a generic name.
The logic of building the path is split between the 2 methods.
We want to give priority to the local image and default to the one from mnoe.

I think I'd rewrite the module like the following, I find it a lot easier to follow.
```ruby
# Add some comments here!
def main_logo_whitebg
  app_image(""mno_enterprise/main-logo-whitebg.png"") || main_logo
end

# Add some comments here!
def main_logo
  logo_path = "mno_enterprise/main-logo.png"
  app_image_path(logo_path) || engine_image_path(logo_path)
end

# Add some comments (eg: get an image from the app etc)
def app_image_path(path)
  app_path = Rails.root.join("/app/assets/images/", path)
  app_path if File.exists?(app_path)
end

def engine_image_path(path)
  engine_path = MnoEnterprise::Engine.root.join("/app/assets/images/", path)
end
```

ouranos:
Ask @x4d3 to review before me once the above is fixed. Thanks

core/app/helpers/mno_enterprise/image_helper.rb
hedudelgado: 👍

core/app/pdf/mno_enterprise/invoice_pdf.rb
hedudelgado: 👍

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+    # Helper method to easily access and select the images
+    # If is_for_invoice is true returns logo appended to the path
+    # If is_for_invoice is false returns logo detached to the path
+    def main_logo_white_bg(is_for_invoice=false)
+      logo = search_main_logo_white_bg
+      is_for_invoice ? logo : "mno_enterprise/#{File.basename(logo)}"
+    end
+
+    # Return the main-logo-whitebg.png if exists, otherwise search main-logo.png
+    def search_main_logo_white_bg
```
x4d3: Remove and inline this method.
Don't use search in this context.

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -0,0 +1,36 @@
+module MnoEnterprise
+  module ImageHelper
+
```
x4d3: add `_path` to all the method so that it's crital clear what they are returning.

- main_logo_white_bg_path
- main_logo_path

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -0,0 +1,36 @@
+module MnoEnterprise
+  module ImageHelper
+    
+    IMAGES_LOCATION = "/app/assets/images/mno_enterprise/"
+    
+    # Helper method to easily access and select the images
+    # If is_for_invoice is true returns logo appended to the path
+    # If is_for_invoice is false returns logo detached to the path
+    def main_logo_white_bg(is_for_invoice=false)
```
x4d3: I think you can rename is_for_invoice to full_path

hedudelgado:
@x4d3 Can you have a look when you have time?

x4d3:
Hello @hedudelgado , did you have the chance to add some specs ?

hedudelgado:
Yes indeed @x4d3 , there is a commit with the tests.

core/spec/helpers/image_helper_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    describe "#main_logo_white_bg_path" do
+      it "should return the engine's main-logo filename, when there are no logos"  do
+        expect(helper.main_logo_white_bg_path).to eq("mno_enterprise/main-logo.png")
+      end
+      
+      it "should return the engine's main-logo full path when there are no logos"  do
+        path = "#{MnoEnterprise::Engine.root}#{IMAGES_LOCATION}/main-logo.png"
+        expect(helper.main_logo_white_bg_path(true)).to eq(path)
+      end
```
ouranos: Some tips from http://www.betterspecs.org
* use contexts to reduce duplication and make it easier to read
* Don't use should
* you're missing a test case

Something along the lines of:
```ruby
context 'with an app main-logo-whitebg' do
   before { #stub file exist for main-logo-whitebg } 
   it "returns the app's main-logo-whitebg"
end
context 'with an app main-logo' do
   before { #stub file exist for main-logo } 
   it "returns the app's main-logo"
end
context 'when there are no logos' do
  it "returns the engine's main-logo filename"
  it "returns the engine's main-logo full path"
end
```
(wording to be improved)

core/spec/helpers/image_helper_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      
+      it "should return the main-logo-whitebg's filename when the file exists"  do
+        allow(File).to receive(:exists?).and_return(true)
+        expect(helper.main_logo_white_bg_path).to eq("mno_enterprise/main-logo-whitebg.png")
+      end
+
+      it "should return the main-logo-whitebg's full path when the file exists"  do
+        path = "#{Rails.root}#{IMAGES_LOCATION}/main-logo-whitebg.png"
+        allow(File).to receive(:exists?).and_return(true)
```
ouranos: `allow(File).to receive(:exists?).with(path).and_return(true)` to make sure we're checking the existence of the correct file

hedudelgado:
Hi @ouranos ,

I have added two new test cases and refactor the specs. Xavier will be off, so you might want to have a look at it instead.

Thanks

PR #314, opened 2017-06-08
[MNOE-307] - Impersonate with the right organization

fgourichon:
@ouranos for review

### DIFF
--- a/api/app/controllers/mno_enterprise/impersonate_controller.rb
+++ b/api/app/controllers/mno_enterprise/impersonate_controller.rb
@@ -17,7 +17,7 @@ module MnoEnterprise
       end
 
       path = mnoe_home_path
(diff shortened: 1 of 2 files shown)


api/app/controllers/mno_enterprise/impersonate_controller.rb
```diff
@@ -15,7 +15,11 @@ def create
       else
         flash[:notice] = "User doesn't exist"
       end
-      redirect_to mnoe_home_path
+
+      path = mnoe_home_path
+      path = add_param_to_fragment(path + "#!/" , 'dhbRefId', params[:dhbRefId]) if params[:dhbRefId].present?
```
ouranos: I'm not sure you need the `+ "!/"` see the [add_param_to_fragment](https://github.com/maestrano/mno-enterprise/blob/d3cddef17bf068e6ef14a20cb79fdcd0e6423d97/core/spec/controllers/mno_enterprise/application_controller_spec.rb#L15-L20)

PR #316, opened 2017-06-09
Ability - case insensitive check of admin_role

ouranos:


core/lib/mno_enterprise/concerns/models/ability.rb
```diff
@@ -141,7 +141,7 @@ def impac_abilities(user)
 
   # Abilities for admin user
   def admin_abilities(user)
-    if user.admin_role == 'admin'
+    if user.admin_role.try(:downcase) == 'admin'
```
x4d3: I would prefer `'admin'.casecmp(user.admin_role)`

ouranos: Ah didn't know this one.

Interestingly, it seems better to use `downcase + ==` in ruby 2.4:
* https://github.com/bbatsov/rubocop/issues/4277
* https://github.com/JuanitoFatas/fast-ruby/issues/123

In anycase, the role is not unicode so it should be safe to use `'admin'.casecmp(user.admin_role).zero?`

ouranos: Updated, let me know if good and I'll squash the commits once the specs are green

PR #326, opened 2017-06-19
v4.0 deprecation

ouranos:
@x4d3 FYI, planning to merge that when starting `v4.0` branch tomorrow


api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
```diff
@@ -35,10 +35,6 @@ def show
   end
 
   def app_relation
-    if MnoEnterprise.marketplace_listing
-      MnoEnterprise::App.where('nid.in' => MnoEnterprise.marketplace_listing)
-    else
-      MnoEnterprise::App.all
-    end
+    MnoEnterprise::App.all
```
BrunoChauvet: :heart_decoration:

api/spec/controllers/mno_enterprise/jpi/v1/marketplace_controller_spec.rb
```diff
@@ -70,108 +70,85 @@ def hash_for_apps(apps)
     describe 'GET #index' do
       subject { get :index }
 
-      context 'when marketplace_listing is set' do
+      before do
+        api_stub_for(get: '/apps', response: from_api([app]))
+        # TODO: Shouldn't need to stub this
+        api_stub_for(get: "/apps/#{app.id}/shared_entities", response: from_api([]))
```
BrunoChauvet: Would it be possible to have a default stub in the `spec_helper.rb`?
```ruby
 api_stub_for(get: /\/apps\/\d+\/shared_entities", response: from_api([]))
```

ouranos: I don't worry too much about this as this is being removed as part of the switch to api v2

ouranos: I originally started all my changes on APIv1 that's why we have this

PR #329, opened 2017-06-23
Remove deprecated Intercom settings

ouranos:


core/lib/mno_enterprise/core.rb
```diff
@@ -195,15 +195,12 @@ def self.mail_adapter=(adapter)
...
-  mattr_accessor :intercom_api_key
-  @@intercom_api_key = nil
-
   mattr_accessor :intercom_token
   @@intercom_token = nil
 
   # Define if Intercom is enabled. Only if the gem intercom is present
   def self.intercom_enabled?
-    defined?(::Intercom) && ((intercom_app_id && intercom_api_key) || intercom_token)
+    defined?(::Intercom) && (intercom_app_id && intercom_token)
```
hedudelgado: I understand before, but now, why not just?
```
defined?(::Intercom) && intercom_app_id && intercom_token
```

ouranos: My bad :sweat_smile:

PR #331, opened 2017-06-23
[MNOE-449] Webstore configuration

ouranos:
**WIP**
@alachaum @BrunoChauvet  @x4d3  TBC but if you want to have a first look

* Settings fetched from the Mnohub tenant API
* Ability to edit and save settings
* Refactor settings structure (frontend settings are nested under `dashboard`)

See the README.md and CHANGELOG.md changes

**TODO**
* Read only flags (`<field>_readonly`)
* Nex! Adapter
* Other tenant fields (domain, address, ...)
* SSL management
* A few ENV vars left (SMTP Settings, ...) 
* App catalog via TenantApp
* Documentation (in progress: UPGRADE & README)
* Refactor settings in frontend (in progress)

**Cleanup**
* remove settings.yml & template


UPGRADING.md
```diff
@@ -41,6 +41,15 @@ This is a simple as `bundle update mno-enterprise`.
 
 See below for upgrade between breaking versions.
 
+### Migrating from v3.3 to v4.0
+
+#### Settings and feature flags
+
+Settings and feature flags are not managed dynamically at the MnoHub level rather than via `settings.yml` files.
```
alachaum: "are now managed" (typo)

UPGRADING.md
```diff
@@ -41,6 +41,15 @@ This is a simple as `bundle update mno-enterprise`.
 
 See below for upgrade between breaking versions.
 
+### Migrating from v3.3 to v4.0
```
alachaum: Let's specify it's ruby 2.3+ only.

core/app/models/mno_enterprise/tenant.rb
```diff
@@ -1,6 +1,55 @@
 module MnoEnterprise
   class Tenant < BaseResource
+    # == Attributes ===========================================================
     property :created_at, type: :time
     property :updated_at, type: :time
+#   property :frontend_config
```
alachaum: Remove? Note that if you do not specify the property you may receive a "NoMethod" error when calling the property when nil.

core/app/models/mno_enterprise/tenant.rb
```diff
@@ -1,6 +1,55 @@
...
+
+    # jsonapi-client build the url with attributes[:primary_key]
```
alachaum: Can you add a line saying: "No primary key required here as this class is designed to hit the /tenant (singular resource) endpoint"

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -0,0 +1,366 @@
+# frozen_string_literal: true
+module MnoEnterprise
+  # Frontend configuration management
+  class TenantConfig
+    # == Constants ============================================================
+    # JSON schema for the configuration
+    # This is used to provide default values and to generate the form in the frontend
+    # This *MUST* be updated any time a new feature flag is added
+    CONFIG_JSON_SCHEMA = {
```
alachaum: We may want to load this from a file eventually....otherwise rubocop is going to complain on the number of lines in this class!

core/app/models/mno_enterprise/tenant_reporting.rb
```diff
@@ -0,0 +1,10 @@
+module MnoEnterprise
+  class TenantReporting < BaseResource
```
alachaum: We may need to specify `self.primary_key = :no_primary_key` - same as Tenant

core/lib/generators/mno_enterprise/install/templates/config/application.yml
```diff
@@ -4,29 +4,5 @@
 tenant_id: my_tenant_id
 tenant_key: my_tenant_access_key
 
-# Emailing platform
-# MANDRILL_API_KEY: changeme
-# SPARKPOST_API_KEY: changeme
-
 # Rails secret key
 SECRET_KEY_BASE: <%= SecureRandom.hex(64) %>
```
alachaum: ❤️  the new config file!

core/lib/mno_enterprise/app_manager.rb
```diff
@@ -0,0 +1,33 @@
+module MnoEnterprise
+  # Abstract the app management logic
+  #
+  class AppManager
```
alachaum: I would rather be in favour of `SystemManager` to not create any confusion with the `App` model.

core/app/models/mno_enterprise/tenant_reporting.rb
ouranos: It's only used for write operations.

ouranos:
Applied changes and rebased
Merging when CI is green

PR #332, opened 2017-06-23
Adapt Admin Controllers to v2

x4d3:


api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -10,20 +10,16 @@ def index
...
+        response.headers['X-Total-Count'] = query.meta.record_count
       end
     end
 
     # GET /mnoe/jpi/v1/admin/organizations/1
     def show
-      @organization = MnoEnterprise::Organization.find(params[:id])
-      @organization_active_apps = @organization.app_instances.active.to_a
+      @organization = MnoEnterprise::Organization.find_one(params[:id], :app_instances, :credit_card)
+      @organization_active_apps = @organization.app_instances.select(&:active?)
```
ouranos: Can we filter with `where` so the filtering is done on the MnoHub side?

core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -9,13 +9,21 @@ class BaseResource < ::JsonApiClient::Resource
...
-      relation.order_by(params[:order_by]) if params[:order_by]
+    def self.apply_query_params(params, relation = self.where)
+      relation.paginate(page: 1 + params[:offset].to_i/params[:limit].to_i, per_page: params[:limit].to_i) if params[:limit] && params[:offset]
+      relation.order(adapt_order_by(params[:order_by])) if params[:order_by]
       relation.where(params[:where]) if params[:where]
       relation
     end
 
+    def self.adapt_order_by(input)
+      split = input.split('.')
```
ouranos:
```ruby
field,order = input.split('.')
if (order == 'desc')
  '-' + field
else
  field
end
```

x4d3: ah yes nice !

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
x4d3: I am not sure nested filtering works. They were already retrieved and loaded.

core/app/models/mno_enterprise/tenant_reporting.rb
```diff
@@ -2,9 +2,14 @@ module MnoEnterprise
   class TenantReporting < BaseResource
     # == Class Methods ========================================================
 
+    # jsonapi-client build the url with attributes[:primary_key]
+    self.primary_key = :no_primary_key
```
ouranos: I don't think that's needed? I think it's only used when doing writes

x4d3:
Yep, most of the api call needs to be  changed to do a fetch_all (marketplace apps for examples)

ouranos:
I've added a change related to what I did on maestrano/maestrano-hub#552

@BrunoChauvet could you review it? (You can review the full PR if you feel like it :innocent: )

core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -8,14 +8,34 @@ class BaseResource < ::JsonApiClient::Resource
...
+        list = list.pages.next
+      end
+      result
+    end
+
     # TODO: Replace limit and offset parameters by page and per_page
-    def self.apply_query_params(relation, params)
-      relation.paginate(page: params[:offset]/params[:limit], per_page: params[:limit]) if params[:limit] && params[:offset]
-      relation.order_by(params[:order_by]) if params[:order_by]
+    def self.apply_query_params(params, relation = self.where)
```
ouranos: Spec?

PR #342, opened 2017-06-30
[MNOE-510] Fix confirmation not being sent

x4d3:
I am not entirely sure it is the best solution.

I am a bit confused by the original implementation from devise and I don't see the case where they manage when email_was is nil. 

Original Implementation
```
        def postpone_email_change_until_confirmation_and_regenerate_confirmation_token
          @reconfirmation_required = true
          self.unconfirmed_email = self.email
          self.email = self.email_was
          self.confirmation_token = nil
          generate_confirmation_token
        end

        def postpone_email_change?
          postpone = self.class.reconfirmable && email_changed? && !@bypass_confirmation_postpone && self.email.present?
          @bypass_confirmation_postpone = false
          postpone
        end
```
Without this change, at user creation, on `resource.save`, `postpone_email_change_until_confirmation_and_regenerate_confirmation_token` is called because `postpone_email_change?` is `true`, `resource.email` is set to `nil` and then validation are failing as the `email` is nil.

The error may be due to the workaround implemented for the callback around saving, emulating Active record behaviour. Maybe it's not in the right order.

@ouranos We could discuss it together.

### DIFF
--- a/core/app/models/mno_enterprise/base_resource.rb
+++ b/core/app/models/mno_enterprise/base_resource.rb
@@ -61,16 +61,16 @@ module MnoEnterprise
     # emulate active record call of callbacks
     def save(*args)
       run_callbacks :save do
-        super()
+        run_callbacks :update do
+          super()
+        end
       end
     end
 
     # emulate active record call of callbacks, a bit different as before_update is called before before_save
     def update_attributes(attrs = {})
       self.attributes = attrs
-      run_callbacks :update do
-        save
-      end
+      save
     end
 
     def cache_key(*timestamp_names)
--- a/core/app/models/mno_enterprise/base_resource.rb
+++ b/core/app/models/mno_enterprise/base_resource.rb
@@ -60,8 +60,9 @@ module MnoEnterprise
 
     # emulate active record call of callbacks
     def save(*args)
+      callback_kind = new_record? ? :create : :update
       run_callbacks :save do
-        run_callbacks :update do
+        run_callbacks callback_kind do
           super()
         end
       end
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -79,6 +79,27 @@ module MnoEnterprise
 
     end
 
+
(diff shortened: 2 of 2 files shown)


core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -61,16 +61,16 @@ def update_attribute(name, value)
     # emulate active record call of callbacks
     def save(*args)
       run_callbacks :save do
```
alachaum: We should differentiate between create and update on the `save` method.

See: https://github.com/rails/rails/blob/310918f6a194bf5752fe1025930881756f5d8a8e/activerecord/lib/active_record/persistence.rb#L128

Based on this we should call the `:save` and `:create` or `:update` callbacks

core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -61,16 +61,16 @@ def update_attribute(name, value)
...
-        super()
+        run_callbacks :update do
+          super()
+        end
       end
     end
 
     # emulate active record call of callbacks, a bit different as before_update is called before before_save
     def update_attributes(attrs = {})
       self.attributes = attrs
```
alachaum: We should consider including `ActiveModel::AttributeAssignment` and use the `assign_attributes` method.
https://github.com/rails/rails/blob/master/activemodel/lib/active_model/attribute_assignment.rb

The current method replaces all the attributes and does not use the model custom setters. Not sure it's ideal.

See original implementation of `update_attributes/update` in `ActiveRecord::Persistence`:
https://github.com/rails/rails/blob/310918f6a194bf5752fe1025930881756f5d8a8e/activerecord/lib/active_record/persistence.rb#L280

x4d3:
@alachaum I've adressed your comment regarding the differentiation between update and create.
I created a Jira regarding assign_attributes

alachaum:
WTF the build is red?

ouranos:
I'm taking this

ouranos:
Closing in favor of #473

PR #349, opened 2017-07-03
[MNOE-474] Api V2 - Adapt Reviews

x4d3:


api/app/controllers/mno_enterprise/jpi/v1/admin/app_answers_controller.rb
```diff
@@ -14,13 +14,14 @@ def create
 
     def app_answer_params
       # for an admin, the organization does not matter
-      organization_id = current_user.organizations.first.id
+      orga_relation = current_user.orga_relations.first
       params.require(:app_answer).permit(:description)
-        .merge(user_id: current_user.id, question_id: parent.id, organization_id: organization_id, app_id: parent.app_id)
+        .merge(reviewer_id: orga_relation.id, reviewer_type: 'OrgaRelation',
+               parent_id: parent.id, reviewable_id: parent.reviewable_id, reviewable_type: 'App')
```
ouranos: Was the API v1 managing the polymorphism on MnoHub side?

api/app/controllers/mno_enterprise/jpi/v1/app_reviews_controller.rb
```diff
@@ -46,49 +42,64 @@ def update
...
-      MnoEnterprise::AppReview
+      Review
+    end
+
+    def initial_scope
+      review_klass
+    end
+
+    def root_params
+      symbol = self.class.name.demodulize.chomp('Controller').underscore.singularize.to_sym
```
ouranos: could you just do:
```ruby
"app_#{review_klass.table_name}".to_sym
```
?

x4d3: Sounds like the idea.

api/app/controllers/mno_enterprise/jpi/v1/admin/app_answers_controller.rb
x4d3: Yes it was. And I don't think it was a good idea. I think it is better to have a clean ReviewResource on the MnoHub side.

image

api/app/controllers/mno_enterprise/jpi/v1/app_reviews_controller.rb
ouranos: [Yes you can!](http://www.rubydoc.info/gems/json_api_client/JsonApiClient/Resource#table_name-class_method)

[We've been  using it already](https://github.com/maestrano/mno-enterprise/blob/4.0/core/app/models/mno_enterprise/tenant.rb#L27-L29)

x4d3: Ah ok :) Fair enough.

api/app/controllers/mno_enterprise/jpi/v1/admin/app_answers_controller.rb
```diff
@@ -14,13 +14,14 @@ def create
...
     end
 
     def parent
-      @parent ||= MnoEnterprise::AppQuestion.find(params[:question_id])
+      @parent ||= MnoEnterprise::Question.find_one(params[:question_id])
```
ouranos: Since the models are no longer namespaced, should we add a `reviewable_type: 'App'` to be safe?

ouranos: Same for all other models

api/app/controllers/mno_enterprise/jpi/v1/app_answers_controller.rb
```diff
@@ -2,21 +2,20 @@ module MnoEnterprise
...
       #do nothing because answers doesn't affect on app rating
     end
 
     def review_klass
-      MnoEnterprise::AppAnswer
+      MnoEnterprise::Answer
+    end
+
+    def initial_scope
+      review_klass.where(parent_id: params[:parent_id])
```
ouranos: Should  we add the `reviewable_type` scoping here too?

api/app/controllers/mno_enterprise/jpi/v1/app_reviews_controller.rb
```diff
@@ -46,49 +42,64 @@ def update
     def destroy
       @app_review.destroy
       after_save
-
+      @app_review = nil
       render :show
```
ouranos: Should we have a `head :no_content` rather than a `render :show`?

x4d3: The frontend is using the apps average rating :( That is why it is uing a show. Which i not working if  you don't set app_review to nil, as the access to the fields are not possible, once destroy has been called.

api/app/controllers/mno_enterprise/jpi/v1/app_answers_controller.rb
x4d3: Yes we could add.

api/app/controllers/mno_enterprise/jpi/v1/admin/app_answers_controller.rb
x4d3: Yes, we can do that.

x4d3: I don't think it is is necessary. As it is a find, it is going to replace a direct questions/id request to a more complicated `where`: questions/filter[id]=id&filter[reviewable_type]=App

api/app/controllers/mno_enterprise/jpi/v1/app_answers_controller.rb
x4d3: No need too, as it is going to be added in the `AppReviewsController` index method anyway.

PR #351, opened 2017-07-05
[MNOE-448] Add DNS/SSL Management

ouranos:
Frontend PR coming soon

### DIFF
--- a/core/lib/mno_enterprise/platform_adapters/local_adapter.rb
+++ b/core/lib/mno_enterprise/platform_adapters/local_adapter.rb
@@ -18,6 +18,16 @@ module MnoEnterprise
         def fetch_assets
           # NOOP
         end
+
+        # @see MnoEnterprise::PlatformAdapters::Adapter#update_domain
+        def update_domain(domain_name)
+          # NOOP
+        end
+
+        # @see MnoEnterprise::PlatformAdapters::Adapter#add_ssl_certs
+        def add_ssl_certs(cert_name, public_cert, cert_bundle, private_key)
+          # NOOP
+        end
       end
     end
   end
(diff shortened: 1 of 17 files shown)


ouranos:
See frontend screenshot at maestrano/mnoe-admin-panel#31

core/lib/mno_enterprise/platform_adapters/local_adapter.rb
```diff
@@ -18,6 +18,16 @@ def publish_assets
         def fetch_assets
           # NOOP
         end
+
+        # @see MnoEnterprise::PlatformAdapters::Adapter#update_domain
+        def update_domain(domain_name)
+          # NOOP
```
x4d3: isn't it to make the `update_domain` method fail in the controller ?

core/lib/mno_enterprise/platform_adapters/nex_adapter.rb
```diff
@@ -33,8 +36,61 @@ def fetch_assets
...
 
+        # @see MnoEnterprise::PlatformAdapters::Adapter#update_domain
+        def update_domain(domain_name)
+          domain = NexClient::Domain.new(cname: domain_name)
+          domain.relationships.origin = nex_app
+          domain.save
+
+          # Display errors if any
+          if domain.errors.any?
+            # display_record_errors(domain)
```
x4d3: I would advocate to return the domain anyway, and display the domain.errors in the controller.

core/lib/mno_enterprise/platform_adapters/nex_adapter.rb
```diff
@@ -33,8 +36,61 @@ def fetch_assets
...
+            false
+          else
+            domain
```
x4d3: not a big fan of the return false of the object pattern.

PR #360, opened 2017-07-06
[MNOE-533] Reload her team object and fix fetch teams

hedudelgado:
Hey @ouranos ,
Adding Jira link since Jira did not get the PR
https://maestrano.atlassian.net/browse/MNOE-533

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/teams_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/teams_controller.rb
@@ -16,7 +16,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::TeamsController
   # GET /mnoe/jpi/v1/organizations/:organization_id/teams
   def index
     authorize! :read, parent_organization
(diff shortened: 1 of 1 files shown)


api/lib/mno_enterprise/concerns/controllers/jpi/v1/teams_controller.rb
```diff
@@ -16,7 +16,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::TeamsController
   # GET /mnoe/jpi/v1/organizations/:organization_id/teams
   def index
     authorize! :read, parent_organization
-    @teams = parent_organization.teams
+    @teams = MnoEnterprise::Team.where(organization_id: params[:organization_id])
```
ouranos: Can we do a `parent_organization.teams.reload` instead?

x4d3: If I remember correctly, the reload was not working. That is why we thought it would be better do a proper load.

PR #361, opened 2017-07-06
Fix V2 Issues

x4d3:


api/app/controllers/mno_enterprise/jpi/v1/admin/invites_controller.rb
```diff
@@ -39,7 +39,7 @@ def send_org_invite(invite)
       # Generate token if not generated
       user.send(:generate_confirmation_token!) if !user.confirmed? && user.confirmation_token.blank?
 
-      MnoEnterprise::SystemNotificationMailer.organization_invite(invite).deliver_later
+      MnoEnterprise::SystemNotificationMailer.organization_invite(invite).deliver_now
```
ouranos: Why remove the `deliver_later`? By default it'll be performed now but if we add sidekiq it'll automatically use it.

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -100,12 +100,13 @@ def user_params
...
+      user.password = Devise.friendly_token
+      user.save!
 
-      # Reset the confirmation field so we can track when the invite is send - #confirmation_sent_at is when the confirmation_token was generated (not sent)
-      # Not ideal as we do 2 saves, and the previous save trigger a call to the backend to validate the token uniqueness
-      user.attributes = {confirmation_sent_at: nil, confirmation_token: nil}
-      user.save
+      # # Reset the confirmation field so we can track when the invite is send - #confirmation_sent_at is when the confirmation_token was generated (not sent)
+      # # Not ideal as we do 2 saves, and the previous save trigger a call to the backend to validate the token uniqueness
+      # user.attributes = {confirmation_sent_at: nil, confirmation_token: nil}
```
ouranos: Why is this commented? Do we still have the same behavior?

x4d3: Well spotted, I was doing some test, and I did not check if it was the same behavior.

api/app/controllers/mno_enterprise/jpi/v1/admin/invites_controller.rb
x4d3: Invite is not serializable. Should we replace it to take invite_id instead ?

PR #364, opened 2017-07-07
[MNOE-477] Monkey patch json_api_client and fix specs

BrunoChauvet:


core/lib/json_api_client_extension/resource_extension.rb
```diff
@@ -0,0 +1,23 @@
+# MonkeyPatching json_api_client, see: https://github.com/chingor13/json_api_client/pull/263
```
x4d3: Could we fork, and point to the gem instead ? Or is it better to monkey patch ? @ouranos what is your take on this ?

ouranos: This is fine for now as a PR is opened upstream.
It's better to get the upstream fixed than depend on a fork

PR #367, opened 2017-07-10
[MNOE-542] Parameterise ActionMailer config

ouranos:
@alachaum what do you think?

I didn't use `NEX_xxx` variables here as I want this part to be platform agnostic. I can set `mailer_default_host` on the Docker entrypoint based on the `NEX_xxx` variables.

Come to think of it, `mailer_default_host|protocol` is not the most accurate name as it's used to generate link (to page and assets) in the emails we send. It's actually referring to the frontend host.

Not sure how to spec that as I can't reinitialise the app in rspec.

(WIP) I'm doing another pass on all the config to make sure nothing is missing


ouranos:
Merging, I'll do another PR with the remaining fixes

PR #375, opened 2017-07-16
[IMPAC-623] create email alert on widget kpi create

xaun:
@ouranos small change to automatically create email alerts for widget KPIs on create - please review, let me know if this change can be specced? As I'm not sure it's possible with api stub jazz.


xaun:
@ouranos spec added, please review

PR #384, opened 2017-07-27
Expose devise config in the frontend

ouranos:
Need to now if the registration are disabled

See maestrano/mno-enterprise#384


alexnoox:
Related to PR https://github.com/maestrano/mnoe-admin-panel/pull/40

PR #385, opened 2017-07-31
[MNOE-558] Filter out rejected comments from feedbacks

adamaziz15:
@ouranos This should take care of the issue regarding rejected comments being included in the response when fetching reviews. Please review.


ouranos:
@alexnoox @adamaziz15 FYI

I've ported maestrano/mno-enterprise-angular#151 in the backend and updated the specs.

alexnoox:
@ouranos The filter in https://github.com/maestrano/mno-enterprise-angular/pull/151 should then be removed.

ouranos:
@alexnoox Yeah I'll remove it in `v2` to avoid any regressions if the backend is not up to date.

PR #392, opened 2017-08-04
[MNOE-477] Add subscription cancellation

BrunoChauvet:


api/app/views/mno_enterprise/jpi/v1/subscriptions/cancel.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.subscription do
```
ouranos: You can get rid of this partial and do a `render :show` in the `cancel` method

ouranos: Same for `create`, `update`

PR #395, opened 2017-08-04
[MNOE-560] Missing Translations

adamaziz15:
@ouranos Please review


adamaziz15:
@ouranos Updated the dummy translations file, it should include all current translations now

PR #407, opened 2017-08-17
[MNOE-570] Storing the time of agreement to TOS

enizor:
The time of agreement of TOS is stored in user[:metadata][tos_accepted_at]
It is then used in the /confirmation to create (or not) the TOS checkbox.
@ouranos

### DIFF
--- a/core/lib/mno_enterprise/concerns/controllers/auth/confirmations_controller.rb
+++ b/core/lib/mno_enterprise/concerns/controllers/auth/confirmations_controller.rb
@@ -92,6 +92,9 @@ module MnoEnterprise::Concerns::Controllers::Auth::ConfirmationsController
     end
 
     if resource.errors.empty?
+      if params[:tos]
+        params[:user][:meta_data] = resource.meta_data.merge(tos_accepted_at: Time.current)
+      end
       resource.assign_attributes(params[:user]) unless resource.confirmed?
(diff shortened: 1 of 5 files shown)


frontend/spec/controllers/mno_enterprise/auth/confirmations_controller_spec.rb
```diff
@@ -20,29 +20,46 @@ module Auth
...
+          
+          context 'when TOS not accepted' do
+            it 'shows the TOS checkbox' do
+              expect(response.body).to include('id="tos"')
+            end
+          end
         end
       end
-    
+
```
ouranos: Remove the extra space

core/lib/mno_enterprise/concerns/controllers/auth/confirmations_controller.rb
```diff
@@ -92,6 +92,9 @@ def finalize
     end
 
     if resource.errors.empty?
+      if params[:tos]
```
ouranos: Should you test `params[:tos] == "accept"`?

enizor: The form can be submitted only when the TOS are accepted so it souldn't be required. 
However, checking the exact string will improve clarity at no cost, I'm adding it.

PR #412, opened 2017-08-18
Admin build: inject impac-angular dependency

ouranos:
Inject impac-angular dependency locked on the yarn.lock version

Used for the dashboard designer
If not dependency is found, it won't inject the resolution (eg: for older version)

### DIFF
--- a/frontend/lib/tasks/mnoe_admin_tasks.rake
+++ b/frontend/lib/tasks/mnoe_admin_tasks.rake
@@ -28,6 +28,29 @@ namespace :mnoe do
       end
     end
 
+    # Override the frontend bower.json with the locked versions
+    def override_admin_dependencies
+      resolved_version = resolve_dependencies
+      unless resolved_version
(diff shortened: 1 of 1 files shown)


ouranos:
I'll open another PR to dry the build between admin and dashboard on the v2

frontend/lib/tasks/mnoe_admin_tasks.rake
```diff
@@ -11,6 +11,7 @@ namespace :mnoe do
     # Default version
     MNOE_ADMIN_PANEL_VERSION = 'v1.0.x'
     MNOE_ADMIN_PANEL_PKG = "git+https://git@github.com/maestrano/mnoe-admin-panel.git##{MNOE_ADMIN_PANEL_VERSION}"
+    IMPAC_ANGULAR_VERSION = 'v1.5.x'
```
cesar-tonnoir: unused constant

PR #416, opened 2017-08-21
I18n Refactoring

ouranos:


api/app/views/mno_enterprise/config/show.js.coffee
```diff
@@ -3,13 +3,14 @@
 angular.module('mnoEnterprise.configuration', [])
   .constant('IMPAC_CONFIG', <%= Settings.impac.to_json.html_safe %>)
   .constant('I18N_CONFIG', {
-    enabled: <%= MnoEnterprise.i18n_enabled %>,
-    available_locales: <%= I18n.available_locales.to_json.html_safe %>
+    enabled: <%= Settings.system.i18n.enabled %>,
+    available_locales: <%= @available_locales.to_json.html_safe %>,
+    preferred_locale: <%= Settings.system.i18n.preferred_locale.to_json.html_safe %>
```
x4d3: maybe introduce `@preferred_local' for consistency ?

core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -130,9 +130,24 @@ def ==(o)
   end
 end
 
+LocaleMiddleware = Struct.new(:app) do
```
x4d3: Maybe you could extract this to lib file ?

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -410,6 +74,19 @@ def self.reconfigure_mnoe!
...
     end
 
+    # Update the JSON schema with values available after initialization
+    # TODO: Refactor to use Proc in the JSON Schema and call them
+    def self.refresh_json_schema!
+      # Re-evaluate the availables locales as the list is only populated after initialization
+      available_locales = I18n.available_locales.map(&:to_s).select {|x| x =~ /[[:alpha:]]{2}-[A-Z]{2}/}
+      locale_map = Hash[available_locales.map {|l| [l, I18n.t('language', locale: l, fallback: false, default: nil) || l] }]
+
+      json_schema['properties']['system']['properties']['i18n']['properties']['available_locales']['items']['enum'] = available_locales
```
x4d3: I would suggest ?
```
props = json_schema['properties']['system']['properties']['i18n']['properties']
props['available_locales']['items']['enum'] = available_locales
props['preferred_locale']['enum'] = available_locales
props['available_locales']['x-schema-form']['titleMap'] = locale_map
props['preferred_locale']['x-schema-form']['titleMap'] = locale_map
```

core/app/models/mno_enterprise/base_resource.rb
alachaum: Let's keep it in this file for now. I prefer to not scatter the logic too much for one-off usages. If we see ourself using another custom middleware then we'll extract.

core/app/models/mno_enterprise/tenant_config.rb
alachaum: @x4d3 note that this section of the code will be refactored anyway. We will use Procs in the schema values to make it dynamic. So we should be able to remove this portion of the code eventually.

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -3,365 +3,26 @@ module MnoEnterprise
...
-                  default: true,
-                  description: "enable frontend configuration from the Admin Panel"
-                }
-              }
-            }
-          }
-        }
-      }
-    }.with_indifferent_access.freeze
+    # See core/config/initializers/00_tenant_config_schema.rb for the configuration
```
alachaum: 👍

core/spec/controllers/mno_enterprise/i18n_spec.rb
```diff
@@ -17,24 +17,39 @@ def index
         I18n.default_locale = :en
       end
 
-      context 'when a valid locale is provided' do
-        it 'sets the provided locale' do
-          get :index, locale: :fr
-          expect(I18n.locale).to eq(:fr)
+      context 'when I18n is disabled' do
+        before { MnoEnterprise.i18n_enabled = false }
```
alachaum: define `subject { I18n.locale }` then use `is_expected.to eq(foo)` in your specs.

core/spec/controllers/mno_enterprise/i18n_spec.rb
```diff
@@ -17,24 +17,39 @@ def index
...
+
+        context 'when a valid locale is provided' do
+          it 'sets the default locale' do
+            get :index, locale: :fr
```
alachaum: You should define `get` in a before action. Then use `it { is_expected.to eq(foo) }` in your specs.

api/app/views/mno_enterprise/config/show.js.coffee
ouranos: Actually if you look at this template, `@available_locales`  is the exception. 
Everything else is coming directly from `Settings`. We have to do some formatting on the`@available_locales` hence the ivar.

core/app/models/mno_enterprise/tenant_config.rb
ouranos: done

core/spec/controllers/mno_enterprise/i18n_spec.rb
ouranos: done

core/spec/controllers/mno_enterprise/i18n_spec.rb
ouranos: done

ouranos:
@alachaum  @x4d3  Code reviewed applied. I'll rebase to fix the conflict

ouranos:
Rebased and added some locales (see last commit).
Waiting for Travis before merging

PR #417, opened 2017-08-21
Public landing/product pages

adamaziz15:


core/lib/generators/mno_enterprise/install/install_generator.rb
```diff
@@ -117,7 +117,7 @@ def notify_about_routes
     mount MnoEnterprise::Engine, at: '/mnoe', as: :mno_enterprise
   end
 
-  root to: redirect('mnoe/auth/users/sign_in')
+  root to: redirect('dashboard/')
```
ouranos: I think we should have a conditional based on the feature flag?
```ruby
if enabled =>  /dashboard/
if not => /sign_in
```
Otherwise, the dashboard start loading for guests before getting redirected to the login page

PR #418, opened 2017-08-23
Forbid Impersonification if access is not allowed

x4d3:


api/spec/controllers/mno_enterprise/impersonate_controller_spec.rb
```diff
@@ -29,30 +28,53 @@ module MnoEnterprise
...
 
           it { is_expected.to redirect_to('/dashboard/#!?dhbRefId=10') }
         end
+
+        context 'when the user does not exist' do
+          before { stub_api_v2(:get, '/users/crappyId', [], %i(deletion_requests organizations orga_relations dashboards user_access_requests)) }
+          subject { get :create, user_id: 'crappyId', dhbRefId: 10 }
+          it do
+            subject
+            is_expected.to redirect_to('/admin/#!?flash=%7B%22msg%22%3A%22User+doesn%27t+exist%22%2C%22type%22%3A%22error%22%7D')
```
ouranos: You could make this a bit nicer with 
```ruby
let(:error_fragment) { "#!?#{URI.encode_www_form([['flash', {msg: "User doesn't exist", type: :error}.to_json]])}" }

...
is_expected.to redirect_to('/admin' + error_fragment)
```

We might want to do a helper :)

core/app/models/mno_enterprise/base_resource.rb
```diff
@@ -36,9 +36,14 @@ def self.adapt_order_by(input)
       end
     end
 
+    def self.find_one!(id, *included)
```
ouranos: This is not used anywhere?
Is it just for the sake of completeness?

x4d3: yes.

api/spec/controllers/mno_enterprise/impersonate_controller_spec.rb
x4d3: True, and we should do one to render it as well.

PR #426, opened 2017-08-31
[MNOE-588] TenantConfig - Allow empty arrays

ouranos:
Bypass Rails `deep_munge` on the Tenant config update to allow empty
array to be saved.


api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
```diff
@@ -71,5 +72,27 @@ def tenant_params
     def tenant_cert_params
       params.require(:tenant).permit(:domain, :certificate, :private_key, :ca_bundle)
     end
+
+    # Bypass Rails `deep_munge` we replace empty arrays by nil:
```
alachaum: `which replaces`

PR #444, opened 2017-09-05
Public Pages - Improve routing and app selection

adamaziz15:
@ouranos As suggested, I added a conditional for routing based on tenant config

### DIFF
--- a/core/app/models/mno_enterprise/tenant_config.rb
+++ b/core/app/models/mno_enterprise/tenant_config.rb
@@ -45,6 +45,8 @@ module MnoEnterprise
       # Reconfigure mnoe
       reconfigure_mnoe!
 
+      update_application_list!
+
       # TODO: update JSON_SCHEMA with readonly fields
       refresh_json_schema!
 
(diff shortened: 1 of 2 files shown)


adamaziz15:
@ouranos Also updated tenant config so that app list is updated to full list after initialization

ouranos:
See #417 for the original PR

core/lib/generators/mno_enterprise/install/install_generator.rb
```diff
@@ -117,7 +117,11 @@ def notify_about_routes
     mount MnoEnterprise::Engine, at: '/mnoe', as: :mno_enterprise
   end
 
-  root to: redirect('dashboard/')
+  if Settings.dashboard.public_pages.enabled
+    root to: redirect('dashboard/')
+  else
+    root to: redirect('mnoe/auth/users/sign_in')
+  end
```
ouranos: This is weird, it's already done in #425
That's the exact same change, I'm surprised that Github is showing it

ouranos:
Incorporated in #458

PR #448, opened 2017-09-06
Add App Catalogue management

ouranos:
* Add App Catalogue management
* Fix marketplace caching to expire when removing app from catalogue


api/config/routes.rb
```diff
@@ -196,6 +196,17 @@
         resources :app_reviews, only: [:index, :show,  :update]
         resources :app_comments, only: [:create]
         resources :app_answers, only: [:create]
+
+        resources :apps, only: [:index] do
+          collection do
+            patch :enable
```
BrunoChauvet: Should the `enable` and `disable` actions be consistent and both available on the collection?

alachaum: We don't yet have the bulk disable action on the API. To be added as a follow up item

PR #465, opened 2017-09-11
[MNOE-602] Add mail invoice specs

hedudelgado:


ouranos:
Closing in favor of #523

PR #469, opened 2017-09-12
Add v2 doc - Feature flags

ouranos:
Still WIP but opening it as it might be useful

@x4d3 FYI


ouranos:
@BrunoChauvet @x4d3 please review
To view it nicely: https://github.com/ouranos/mno-enterprise/tree/feature/doc-v2/docs

I'll add more developer doc shortly so do not merge

PR #473, opened 2017-09-13
[MNOE-510][MNOE-612] Fix devise regression

ouranos:
Replaces #342 

* confirmation email not sent
* password change


api/lib/mno_enterprise/concerns/controllers/jpi/v1/current_users_controller.rb
```diff
@@ -54,6 +54,8 @@ def register_developer
   def update_password
     @user = current_user
     @user = @user.update_password(data: {attributes: password_params}).first
+    # TODO: [APIv2] This is not working!
+    # update_password returns an empty array in case of error
```
BrunoChauvet: Should this be addressed in a separate issue?

ouranos: Yes, it's part of [MNOE-602].
I left the comment there for grep'ing :)

PR #477, opened 2017-09-13
[MNOE-600] Fix size logo on invoice

hedudelgado:


core/app/pdf/mno_enterprise/invoice_pdf.rb
```diff
@@ -138,7 +138,7 @@ def add_page_header
       @pdf.repeat :all do
         @pdf.bounding_box([0, @pdf.bounds.top+@format[:header_size]], width: 540, height: @format[:footer_size]) do
           @pdf.float do
-            @pdf.image image_path('mno_enterprise/main-logo.png'), scale: 0.5
```
x4d3: what happens if the image is bigger, isn't it possible to scale it to a fixed size ?

hedudelgado: Hi @x4d3 . That's true. I have added a new commit. Now the logo will be fitted within 1/3 of the header height and 1/4 of the header width. So the logo won't lose its proportions and will fit perfectly regardless its size.

hedudelgado: https://maestrano.atlassian.net/browse/MNOE-600

Pictures in Jira

PR #480, opened 2017-09-14
[MNOE-617] Expose Organization freeze/unfreeze actions

BrunoChauvet:


core/lib/mno_enterprise/concerns/models/organization.rb
```diff
@@ -9,7 +9,8 @@ module MnoEnterprise::Concerns::Models::Organization
   included do
     custom_endpoint :app_instances_sync, on: :member, request_method: :get
     custom_endpoint :trigger_app_instances_sync, on: :member, request_method: :post
-
+    custom_endpoint :freeze, on: :member, request_method: :patch
+    custom_endpoint :unfreeze, on: :member, request_method: :patch
```
fgourichon: Out of curiosity, why is it a PUT on mnoe, and a PATCH on mnohub?

BrunoChauvet: I'm following the existing conventions per project. It would be preferable to use `PATCH` as it is a partial update of the resource.

PR #484, opened 2017-09-14
[MNOE-611] Inactive Apps should not be exposed on API

x4d3:


BrunoChauvet:
@x4d3 if an app has been launched by a user and is then made inactive, does it still display on the user front-end?

x4d3:
Yes, it should as it's coming from the AppInstance.

PR #499, opened 2017-09-26
[MNOE-674] - Fixed invitation for existing users

fgourichon:


core/spec/models/mno_enterprise/orga_invite_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+    end
+
+    describe '#expired?' do
+      subject { orga_invite.expired? }
+
+      context 'invitation is cancelled' do
+        let(:orga_invite) { FactoryGirl.build(:orga_invite, status: 'cancelled') }
+
+        it { is_expected.to be true }
```
BrunoChauvet: You can use the matchers `be_truthy` and `be_falsey` for better formatting of the specs when running

fgourichon: Done!

PR #500, opened 2017-09-26
Fix Terms of Use not displaying

hedudelgado:


api/lib/mno_enterprise/concerns/controllers/pages_controller.rb
```diff
@@ -64,7 +64,7 @@ def terms
     ts = MnoEnterprise::App.order(updated_at: :desc).select(:updated_at).first.updated_at
     @apps = if ts
               Rails.cache.fetch(['pages/terms/app-list', ts]) do
-                MnoEnterprise::App.order(name: :asc).reject { |i| i.terms_url.blank? }
+                MnoEnterprise::App.fetch_all.reject { |i| i.terms_url.blank? }
```
BrunoChauvet: You should be able to query something like
```ruby
MnoEnterprise::App.fetch_all(MnoEnterprise::App.where(scope: 'terms_url.none'))
```
This will avoid retrieving all the apps from the API to then filter them in memory.
See implementation: https://github.com/maestrano/maestrano-hub/blob/develop/app/resources/base_resource.rb#L65

hedudelgado: Heys @BrunoChauvet , I have sent you an email about this. Please, get back to me :)

PR #501, opened 2017-09-26
[MNOE-582] Remove impersonation request if user modify account setting to authorise admin to access its account

x4d3:


api/lib/mno_enterprise/concerns/controllers/jpi/v1/user_access_requests_controller.rb
```diff
@@ -23,6 +23,18 @@ def last_access_request
 
   # POST /mnoe/jpi/v1/user_access_requests
   def create
+    access_duration = params.require(:access_duration)
+    user = current_user.load_required(:user_access_requests, :'user_access_requests.requester')
+    user.user_access_requests.each do |request|
```
BrunoChauvet: I would move this logic inside the class `UserAccessRequest` as a class method and then call something like
```ruby
MnoEnterprise::UserAccessRequest.approve_pending_requests(user)
```

api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
```diff
@@ -211,6 +211,21 @@ def access_approved(user_access_request_id)
     )
   end
 
+  def access_approved_all(user_id, requester_id, access_duration)
```
BrunoChauvet: Is it possible to pass the objects `user` and `requester` as they have already been loaded before?

api/spec/controllers/mno_enterprise/jpi/v1/user_access_requests_controller_spec.rb
```diff
@@ -33,6 +33,19 @@ module MnoEnterprise
...
+      let!(:stubs) {
+        [
+          stub_api_v2(:post, "/user_access_requests", user_access_request),
+          stub_api_v2(:get, "/users/#{user.id}", user, [:user_access_requests, :'user_access_requests.requester']),
+          stub_api_v2(:get, "/user_access_requests/#{user_access_request.id}", user_access_request, [:requester])
+        ]
+      }
+      subject { post :create, access_duration: 'UNTIL_REVOKED'}
+      it_behaves_like 'jpi v1 protected action'
+    end
```
BrunoChauvet: Can you add a scenario where there are pending `user_access_requests`?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/user_access_requests_controller.rb
x4d3: Ok

api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
x4d3: I don't trust serialization on BaseResource Object.

x4d3:
@BrunoChauvet I've applied your comments.

PR #507, opened 2017-09-28
[MNOE-656] Disable onboarding

hedudelgado:


hedudelgado:
https://github.com/maestrano/mno-enterprise-angular/pull/349

PR #509, opened 2017-09-28
[MNOE-655] Add subscription locales and provisioning data field

hedudelgado:


hedudelgado:
https://github.com/maestrano/mno-enterprise-angular/pull/351

PR #512, opened 2017-10-02
[MNOE-428][MNOE-634]  Rehabilitate Specs

x4d3:


api/spec/controllers/mno_enterprise/jpi/v1/admin/app_instances_controller_spec.rb
```diff
@@ -4,30 +4,35 @@
...
-    before { skip }
 
     render_views
     routes { MnoEnterprise::Engine.routes }
-    before { request.env["HTTP_ACCEPT"] = 'application/json' }
+    before { request.env['HTTP_ACCEPT'] = 'application/json' }
+    before { stub_audit_events }
 
     let(:user) { build(:user, :admin, :with_organizations) }
+    let!(:current_user_stub) { stub_user(user) }
```
ouranos: Why the`let!` and not just:
```
before do
  stub_user(user)
  sign_in user
end
```

ouranos: Same for the other spec files

api/spec/controllers/mno_enterprise/jpi/v1/admin/cloud_apps_controller_spec.rb
```diff
@@ -3,102 +3,60 @@
...
-          expect(app.terms_url).to eq(params[:terms_url])
-        end
-
-        it 'only updates authorized fields' do
-          subject
-          expect(app.name).to eq('My App')
-        end
-
-        it 'is successful' do
+        it do
```
ouranos: You should add a description to your example

api/spec/controllers/mno_enterprise/jpi/v1/admin/cloud_apps_controller_spec.rb
```diff
@@ -3,102 +3,60 @@
...
         it_behaves_like 'a jpi v1 admin action'
-
-        it 'regenerates the API key' do
-          expect_any_instance_of(MnoEnterprise::App).to receive(:regenerate_api_key!)
+        it do
           subject
-        end
-      end
-
-      describe "#refresh_metadata" do
```
ouranos: Why has this example been removed?

api/spec/controllers/mno_enterprise/jpi/v1/admin/cloud_apps_controller_spec.rb
ouranos: Also you're not testing the full behavior.
Missing a spec on the strong params

api/spec/controllers/mno_enterprise/jpi/v1/marketplace_controller_spec.rb
```diff
@@ -117,89 +123,55 @@ def hash_for_apps(apps)
...
           end
-
           context 'if it is stale' do
             let(:last_modified) { app.updated_at - 10.minutes }
             it { is_expected.to have_http_status(:ok) }
           end
         end
       end
-
-      context 'without apps' do
```
ouranos: Why is this example removed?
It's quite important as it's a regression test. The bug appeared on v1 and was fixed but got reintroduced on v2 since the specs where skipped

api/spec/controllers/mno_enterprise/jpi/v1/admin/cloud_apps_controller_spec.rb
x4d3: I should have removed the method too. The MnoHub endpoint has not been implemented.

api/spec/controllers/mno_enterprise/jpi/v1/marketplace_controller_spec.rb
x4d3: I thought it was no more releveant with the introduction of the last_modified on the tenant: https://github.com/maestrano/mno-enterprise/blob/4.0/api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb#L28

api/spec/controllers/mno_enterprise/jpi/v1/admin/app_instances_controller_spec.rb
x4d3: For multiple reasons

- to be sure it was unique
- eventually to be tested (it has been called)
- to be more semantic

x4d3:
@ouranos I've addressed your comments.

PR #519, opened 2017-10-05
[MNOE-665] Handle RO feature flags

ouranos:
Feature flags can be marked as ReadOnly on the MnoHub side.
If a the config contains `flag_readonly: true`, the flag is considered
as readonly and marked as such in the schema.

### DIFF
--- a/core/app/models/mno_enterprise/tenant_config.rb
+++ b/core/app/models/mno_enterprise/tenant_config.rb
@@ -45,8 +45,8 @@ module MnoEnterprise
       # Reconfigure mnoe
       reconfigure_mnoe!
 
-      # TODO: update JSON_SCHEMA with readonly fields
-      refresh_json_schema!
+      # Update JSON_SCHEMA with readonly flags (`<field>_readonly`)
+      refresh_json_schema!(frontend_config)
 
       # # Save settings in YAML format for easy debugging
       # Rails.logger.debug "Settings loaded -> Saving..."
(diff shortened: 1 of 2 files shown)


core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -121,6 +113,39 @@ def self.build_object(schema)
       end
     end
 
+    # Flag field as readonly in the config schema
+    # A field is readonly if the `<field>_readonly` key is `true` in the `frontend_config`
+    def self.update_readonly_fields(schema, frontend_config)
+      frontend_config.each do |current_key, value|
+        if current_key =~ /_readonly$/
+          # Flag field as RO
+          schema['properties'][current_key.gsub('_readonly', '')].merge!('readonly' => true)
```
x4d3: why not ?
`schema['properties'][current_key.gsub('_readonly', '')]['readonly'] = true`

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -121,6 +113,39 @@ def self.build_object(schema)
       end
     end
 
+    # Flag field as readonly in the config schema
+    # A field is readonly if the `<field>_readonly` key is `true` in the `frontend_config`
+    def self.update_readonly_fields(schema, frontend_config)
+      frontend_config.each do |current_key, value|
+        if current_key =~ /_readonly$/
```
x4d3: I would suggest `current_key.end_with?("_readonly")`

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -81,18 +81,10 @@ def self.reconfigure_mnoe!
 
     # Update the JSON schema with values available after initialization
     # TODO: Refactor to use Proc in the JSON Schema and call them
```
alachaum: You can remove this TODO as you've broken down the method.

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -121,6 +113,39 @@ def self.build_object(schema)
       end
     end
 
+    # Flag field as readonly in the config schema
+    # A field is readonly if the `<field>_readonly` key is `true` in the `frontend_config`
+    def self.update_readonly_fields(schema, frontend_config)
```
alachaum: I'd recommend changing the method name to `mark_readonly_fields` or `add_readonly_flags` or `add_readonly_properties`

The name `update_readonly_fields` sounds weird for some reasons 😃

core/app/models/mno_enterprise/tenant_config.rb
alachaum: Or you can do this 😎 
```ruby
if field_label = current_key[/(.*)_readonly$/, 1]
  schema['properties'][field_label]['readonly'] = true
end
```

PS: rubocop recommends assignments rather than merge

ouranos:
@alachaum @x4d3 changes applied

alachaum:
👍

PR #525, opened 2017-10-10
[MNOE-658] Disallow Staff impersonation

ouranos:


api/app/controllers/mno_enterprise/impersonate_controller.rb
```diff
@@ -11,7 +11,10 @@ def create
       session[:impersonator_redirect_path] = params[:redirect_path].presence
       @user = MnoEnterprise::User.find_one(params[:user_id], :deletion_requests, :organizations, :orga_relations, :dashboards, :teams, :user_access_requests)
       unless @user.present?
-        return redirect_with_error("User doesn't exist")
+        return redirect_with_error('User does not exist')
+      end
+      if @user.admin_role.present?
```
fgourichon: Can't admin_role contain 'admin' ?

ouranos: No, we wan't to completely disable impersonation of user having access to the admin panel.
(admin_role = staff or admin, ...)

PR #527, opened 2017-10-11
Fix TenantConfig Array default handling

ouranos:
- Fix Array default
- Fix JSON Schema


ouranos:
@BrunoChauvet could you review the commit I pushed after @x4d3 review? Thanks

ouranos:
Actually wait a bit, I need to fix something else

ouranos:
@BrunoChauvet ready for review

PR #529, opened 2017-10-13
[MNOE-669] Organisation deletion

enizor:
Implements organisation deletion. Main task left is correctly freezeing the org in the backend (last bill, etc)

### DIFF
--- a/api/app/views/mno_enterprise/jpi/v1/current_users/show.json.jbuilder
+++ b/api/app/views/mno_enterprise/jpi/v1/current_users/show.json.jbuilder
@@ -31,10 +31,11 @@ json.cache! ['v2', @user.cache_key] do
     # Embed association if user is persisted
     if @user.id
       json.organizations do
-        json.array! (@user.organizations.select(&:active?) || []) do |o|
+        json.array! @organizations do |o|
           json.id o.id
           json.uid o.uid
           json.name o.name
+          json.active o.active?
           json.currency o.billing_currency
           json.current_user_role @user.role(o)
           json.has_myob_essentials_only o.has_myob_essentials_only
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
@@ -140,6 +140,23 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::OrganizationsController
     render 'members'
   end
 
+
+  # POST /mnoe/jpi/v1/organizations/:id/freeze
+  # Avoid conflicts with #freeze
+  def freeze_org
+    # Check the password provided
+    return head :forbidden unless current_user.class.authenticate_user({email: current_user.email, password: params[:password]})
+    del_req = MnoEnterprise::DeletionRequest.create(deletable_type: "Organization", deletable_id: params[:id], reason: params[:reason])
+    # Processes it
+    del_req.freeze
+    # Refresh cache
+    current_user.refresh_user_cache()
+
+    @organization = organization.load_required(:users, :orga_invites, :orga_relations)
+
+    render json: del_req
+  end
+
   protected
   def member
     @member ||= begin
--- a/core/app/pdf/mno_enterprise/invoice_pdf.rb
+++ b/core/app/pdf/mno_enterprise/invoice_pdf.rb
@@ -253,12 +253,20 @@ module MnoEnterprise
       @pdf.indent(5) do
         @pdf.font_size(8) do
           @pdf.text "<color rgb='999999'> #{t('charging_details.bank_statement', invoice_reference: @data[:invoice_reference])}</color>", inline_format: true
+
           @pdf.text "<color rgb='999999'> #{t('charging_details.currency', currency_name: @data[:invoice_currency_name], currency: @data[:invoice_currency])}</color>", inline_format: true
           if @data[:invoice_fully_paid]
             @pdf.text "<color rgb='999999'>  #{t('charging_details.no_payments')}</color>", inline_format: true
           else
             @pdf.text "<color rgb='999999'>  #{t('charging_details.with_payment', charge_date: @data[:period_charge_date].strftime("%B,%e %Y"), platform_name: MnoEnterprise.app_name)}</color>", inline_format: true
           end
+        end
+      end
+
+      if @invoice.will_org_be_deleted?
+        @pdf.move_down 60
+        @pdf.font_size(20) do
+          @pdf.text "<color rgb='#f21414'> This is the final invoice before account deletion</color>", inline_format: true, style: :bold, align: :center
         end
       end
 
--- a/core/lib/mno_enterprise/concerns/models/organization.rb
+++ b/core/lib/mno_enterprise/concerns/models/organization.rb
@@ -70,6 +70,11 @@ module MnoEnterprise::Concerns::Models::Organization
     !self.account_frozen
   end
 
+  # Visible organizations for the dashboard - active orgs (not frozen) and orgs being frozen
+  def visible?(user)
+    self.active? || (self.freezing && user.role_from_id(self.id) == "Super Admin")
+  end
+
   def payment_restriction
     metadata && metadata['payment_restriction']
   end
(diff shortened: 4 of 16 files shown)


api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
```diff
@@ -168,6 +168,17 @@ def remove_member
     render 'members'
   end
 
+
+    # POST /mnoe/jpi/v1/organizations/:id/freeze
+    def halt
+      # Check the password provided
+      return head :forbidden unless current_user.class.authenticate_user({email: current_user.email, password: params[:password]})
```
ouranos: You can user `current_user.valid_password?(params.require(:password))` instead

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
```diff
@@ -168,6 +168,17 @@ def remove_member
     render 'members'
   end
 
+
+    # POST /mnoe/jpi/v1/organizations/:id/freeze
+    def halt
```
ouranos: Comment is wrong (freeze/halt)
Not sure if `/organization/:id/halt` is the best naming here.

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
```diff
@@ -168,6 +168,17 @@ def remove_member
...
+      # Check the password provided
+      return head :forbidden unless current_user.class.authenticate_user({email: current_user.email, password: params[:password]})
+      del_req = MnoEnterprise::DeletionRequest.create(deletable_type: "Organization", deletable_id: params[:id], reason: params[:reason])
```
ouranos: @x4d3 could you comment on what the best syntax to have here regarding associations?

core/app/pdf/mno_enterprise/invoice_pdf.rb
```diff
@@ -34,6 +34,7 @@ def initialize(invoice)
         @data["customer_#{detail}".to_sym] = invoicable.respond_to?(detail) ? invoicable.send(detail) : nil
       end
 
+      @data[:deletion_notice] = invoicable.account_frozen && !invoicable.frozen_at.nil? && invoicable.frozen_at < invoice.ended_at
```
ouranos: Could extract this to a separate helper method?

core/lib/mno_enterprise/concerns/models/organization.rb
```diff
@@ -67,6 +69,18 @@ def active?
     !self.account_frozen
   end
 
+  def freezing?
```
ouranos: Add a comment to explain what's going on here

core/lib/mno_enterprise/concerns/models/organization.rb
```diff
@@ -67,6 +69,18 @@ def active?
...
+    if self.frozen_at
+      return self.last_invoice_date < self.frozen_at
+    else
+      return false
+    end
+  end
+
+  def visible?(user)
```
ouranos: Same

ouranos (commented):
See comments.
This need to be spec'd

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
enizor: `valid_password?` is defined by `Devise::Models::DatabaseAuthenticatable` which cannot be used here. It could be added in `remote_authenticatable` via a custom endpoint but I'm not sure it is necessary

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
enizor: `freeze` is reserved by Ruby si I had to choose another name. If someone has a better name I'm all for it.

core/app/pdf/mno_enterprise/invoice_pdf.rb
enizor: Added a method for `invoice`: `will_org_be_deleted?`... Couldn't come up with a better name (maybe just `deletion_notice?` - the PDF context is unclear)

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
x4d3: it should be
```
deletion_request = MnoEnterprise::DeletionRequest.new(reason: params[:reason])
deletion_request.relationships.deletable = Organization.find_one(params[:id])
deletion_request.save
```

enizor: `deletion_request.relationships.deletable = Organization.find_one(params[:id])` does not work as it sets `delatable_type` to `"organizations"` instead of `"Organization"`

x4d3: This is odd, as thatthis is what we did for SubscriptionController and it worked.
relationships.organization = MnoEnterprise::Organization.new(id: parent_organization.id)
I would think that we need to update mno_hub to support having deletable_type to "organizations"

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
x4d3: it's reserved for active record, but you can still have an endpoint called freeze. For example : https://github.com/maestrano/maestrano-hub/blob/develop/app/controllers/api/mnoe/v2/organizations_controller.rb#L18

api/lib/mno_enterprise/concerns/controllers/jpi/v1/organizations_controller.rb
x4d3: The deletable relation is indeed missing on the DeletionRequest. I will update it.

enizor:
I can't seem to make the `system_notification_mailer_spec` work. I need to be able to access the organization from the invoice to generate the PDF, and the Factory creates an attribute [:organization] but not a method #organization. In the other cases I worked around it via `allow(invoice).to receive(:organization).and_return(organization)` but I cannot do it when the spec calls `subject.send_invoice(user.id, invoice.id).deliver_now`

enizor:
`allow_any_instance_of` solved all the problems!

enizor:
Is there anything else to change?

x4d3:
Could you please rebase your changes to fix the conflicts ?

enizor:
Rebased

core/lib/mno_enterprise/concerns/models/organization.rb
```diff
@@ -70,6 +70,11 @@ def active?
     !self.account_frozen
   end
 
+  # Visible organizations for the dashboard - active orgs (not frozen) and orgs being frozen
+  def visible?(user)
+    self.active? || self.freezing && user.role_from_id(self.id) == "Super Admin"
```
x4d3: why only the Super Admin can see the Organizations ?

api/app/views/mno_enterprise/jpi/v1/current_users/show.json.jbuilder
```diff
@@ -31,14 +31,17 @@ json.cache! ['v2', @user.cache_key] do
...
       json.organizations do
-        json.array! (@user.organizations.select(&:active?) || []) do |o|
-          json.id o.id
-          json.uid o.uid
-          json.name o.name
-          json.currency o.billing_currency
-          json.current_user_role @user.role(o)
-          json.has_myob_essentials_only o.has_myob_essentials_only
-          json.financial_year_end_month o.financial_year_end_month
+        json.array! @user.organizations do |o|
```
x4d3: I think it would be worth doing the filtering on the controller.
I would suggest to stop using the organization include in current_user (which is bringing issue as it's not paginated anyway) and introduce `@organizations = MnoEnterprise::Organization.where('users.id': current_user.id) and then filter the array on the .visible? method 
Or better, find a way to introduce 'OR' operator in MnoHub Api v2.

x4d3 (commented):
Some small comments.

core/lib/mno_enterprise/concerns/models/organization.rb
enizor: Indeed only the "Super Admin" can see the freezing orgs. I will add parenthesis

enizor:
Changed accordingly

x4d3 (approved):

BrunoChauvet (approved):

PR #535, opened 2017-10-17
[MNO-695] Api v2 Dashboard missing dependency for KPI Alert Recipients

x4d3:
- Include Recipients in Alerts
- Use update_recipients instead of recipient_id

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
@@ -26,8 +26,8 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::KpisController
 
     begin
       response = MnoEnterprise::ImpacClient.send_get('/api/v2/kpis', attrs, basic_auth: auth)
-      # TODO check there was no error, something like
-      # return render json: { message: "Unable to retrieve kpis from Impac API | Error #{response.code}" } unless response.success?
+        # TODO check there was no error, something like
+        # return render json: { message: "Unable to retrieve kpis from Impac API | Error #{response.code}" } unless response.success?
     rescue => e
       return render json: { message: "Unable to retrieve kpis from Impac API | Error #{e}" }
     end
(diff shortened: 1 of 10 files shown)


api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
```diff
@@ -39,11 +39,11 @@ def index
         kpi.merge(
           name: "#{kpi[:name]} #{watchable.capitalize unless kpi[:name].downcase.index(watchable)}".strip,
           watchables: [watchable],
-          target_placeholders: {watchable => kpi[:target_placeholders][watchable]},
+          target_placeholders: { watchable => kpi[:target_placeholders][watchable] },
         )
       end
     end
-    .flatten
+             .flatten
```
ouranos: ?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
```diff
@@ -26,8 +26,8 @@ def index
 
     begin
       response = MnoEnterprise::ImpacClient.send_get('/api/v2/kpis', attrs, basic_auth: auth)
-      # TODO check there was no error, something like
-      # return render json: { message: "Unable to retrieve kpis from Impac API | Error #{response.code}" } unless response.success?
+        # TODO check there was no error, something like
```
ouranos: ?

core/app/models/mno_enterprise/alert.rb
```diff
@@ -2,5 +2,22 @@ module MnoEnterprise
   class Alert < BaseResource
     property :created_at, type: :time
     property :updated_at, type: :time
+    property :kpi_id
```
ouranos: Should the `kpi_id` be exposed? I thought we were getting rid of the foreign keys

x4d3: It should for now, but will be addressed later in https://github.com/maestrano/mno-enterprise/pull/560

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
x4d3: Rubymine...

PR #542, opened 2017-10-20
Fix terms of use not being displayed

hedudelgado:


hedudelgado:
@ouranos

api/lib/mno_enterprise/concerns/controllers/pages_controller.rb
```diff
@@ -65,7 +65,9 @@ def terms
     ts = MnoEnterprise::App.order_by("updated_at.desc").first.try(:updated_at)
     @apps = if ts
       Rails.cache.fetch(['pages/terms/app-list', ts]) do
-        MnoEnterprise::App.order_by("name.ac").reject{|i| i.terms_url.blank?}
+        # Temp solution as translated fields can not be filtered or sorted
+        # MnoEnterprise::App.order_by("name.ac").reject{|i| i.terms_url.blank?}
+        MnoEnterprise::App.all.reject { |i| i.terms_url.blank? }
```
ouranos: You should then do the sorting in ruby, otherwise the order will be wrong.
I'm surprised the specs haven't broken so please have a look at them to see if they need to be updated.

hedudelgado: Hi @ouranos , I have updated the PR. Thanks

api/lib/mno_enterprise/concerns/controllers/pages_controller.rb
```diff
@@ -65,7 +65,9 @@ def terms
     ts = MnoEnterprise::App.order_by("updated_at.desc").first.try(:updated_at)
     @apps = if ts
       Rails.cache.fetch(['pages/terms/app-list', ts]) do
-        MnoEnterprise::App.order_by("name.ac").reject{|i| i.terms_url.blank?}
+        # Temp solution as translated fields can not be filtered or sorted
+        # MnoEnterprise::App.order_by("name.ac").reject{|i| i.terms_url.blank?}
+        MnoEnterprise::App.all.reject { |i| i.terms_url.blank? }.sort! { |a,b| a.name.downcase <=> b.name.downcase }
```
x4d3: why not `sort_by{|h|h.name.downcase}`

hedudelgado: Sure, like yours more, 0 advantages of my approach compare to yours.
Updated.

PR #543, opened 2017-10-20
Fix terms of use not being displayed

hedudelgado:


hedudelgado:
@ouranos

ouranos:
Fixed in `3.3` by #542

PR #552, opened 2017-10-30
[MNOE-687] Fix Csv Upload Issue + Add external_id to organization

x4d3:


ouranos:
Is the CSV bug tracked anywhere?

PR #556, opened 2017-11-01
[MNOE-712] Add missing locale

BrunoChauvet:


core/config/locales/templates/dashboard/provisioning/en.yml
```diff
@@ -70,11 +70,11 @@ en:
             status: "Status"
             states:
               requested: requested
-              approved: approved
```
ouranos: No longer a need for approved?

BrunoChauvet: The `approved` state has been removed, the transition is now from called `approve` and transition from `requested` to `provisioning`

PR #567, opened 2017-11-03
[IMPAC-665] Sync status on impac dashboard

enizor:
The (trigger_)app_instance_sync will also return has_running_cube
for a awared 'Real-time syncing' on Impac-Angular
Depends on https://github.com/maestrano/maestrano-hub/pull/802

### DIFF
--- a/api/spec/controllers/mno_enterprise/jpi/v1/app_instances_sync_controller_spec.rb
+++ b/api/spec/controllers/mno_enterprise/jpi/v1/app_instances_sync_controller_spec.rb
@@ -9,7 +9,6 @@ module MnoEnterprise
     routes { MnoEnterprise::Engine.routes }
     before { request.env['HTTP_ACCEPT'] = 'application/json' }
 
-
     #===============================================
     # Assignments
     #===============================================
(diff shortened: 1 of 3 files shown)


enizor:
Je dois réparer les specs associées

api/spec/controllers/mno_enterprise/jpi/v1/app_instances_sync_controller_spec.rb
```diff
@@ -38,7 +38,7 @@ module MnoEnterprise
       HashWithIndifferentAccess.new(name: 'a_name', status: 'FAILED', date: nil)
     ] }
 
-    let!(:organization_with_connectors) { build(:organization, connectors: connectors) }
+    let!(:organization_with_connectors) { build(:organization, connectors: connectors, has_running_cube: false) }
```
cesar-tonnoir: Could worth adding a short test case for when `org.has_running_cube = true`: you can make sure `has_running_cube` is returned by the controller, so that if we break it with a later change, the spec will turn red..

enizor:
added specs

PR #571, opened 2017-11-08
[MNOE-783] Fix registration route not being loaded

ouranos:
Fix an issue when the registration is disabled in the settings.yml and
then enabled via the MnoHub settings.
Since the routes are being build based on the User#devise_modules which
is built when loading the class, the registration route is never added.

The Devise::Models::Registerable is only adding a class method to the
user model (.new_with_session), so it's ok to always add it.

### DIFF
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -53,11 +53,10 @@ module MnoEnterprise
     end
 
     devise_modules = [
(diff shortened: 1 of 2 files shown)


ouranos:
"In progress" as I'll do more tests tomorrow to check if there are any unintended side effects.
Happy so far with my testing.

ouranos:
@fgourichon @x4d3 rebased and happy with testing. Please review.

PR #577, opened 2017-11-14
[MNOE-776] Ability to update_member and remove_member from admin

manu-d:
Needed PUT #update_member and PUT #remove_member actions in admin controller for organizations

=> Move update_member logic in organization model
=> Create shared example for these two methods in a new file

### DIFF
--- a/api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
+++ b/api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
@@ -136,10 +136,8 @@ module MnoEnterprise
 
     # PUT /mnoe/jpi/v1/admin/organizations/:id/update_member
     def update_member
-      role = params.require(:member).require(:role)
       @organization = MnoEnterprise::Organization.find_one(params[:id], :users, :orga_invites, :orga_relations)
-      user = requested_user_in_org(@organization)
-      @organization.update_user_role(current_user, user, role)
+      @organization.update_user_role(current_user, requested_user, params.require(:member).require(:role))
 
       render 'members'
     end
@@ -147,7 +145,7 @@ module MnoEnterprise
     # PUT /mnoe/jpi/v1/admin/organizations/:id/remove_user
     def remove_member
       @organization = MnoEnterprise::Organization.find_one(params[:id], :users, :orga_invites, :orga_relations)
-      user = requested_user_in_org(@organization)
+      user = requested_user
       if user.is_a?(MnoEnterprise::User)
         @organization.remove_user!(user)
       elsif user.is_a?(MnoEnterprise::OrgaInvite)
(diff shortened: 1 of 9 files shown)


api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -188,6 +211,15 @@ def user_params
       params.require(:user).permit(:email, :name, :surname, :phone)
     end
 
+    def requested_user_in_org(organization)
+      user = organization.users.find { |u| u.id == params.require(:member)[:id] }
```
x4d3: could you please add some comment on why you try first with id and then by email ?

core/lib/mno_enterprise/concerns/models/organization.rb
```diff
@@ -85,6 +85,40 @@ def orga_relation(user)
     }
   end
 
+  # Update a user role within self on behalf on user_updating
+  #
+  # user_updating - A MnoEnterprise::User who performed the action
+  # user_to_update - The user to update. ∈ [MnoEnterprise::User, MnoEnterprise::OrgaInvite]
+  # new_role - The role to assign to user_to_update
+  #
+  def update_user_role(user_updating, user_to_update, new_role)
```
x4d3: I would suggest `user, updated_user`

core/lib/mno_enterprise/concerns/models/organization.rb
```diff
@@ -85,6 +85,40 @@ def orga_relation(user)
     }
   end
 
+  # Update a user role within self
+  #
+  # user - A MnoEnterprise::User who performed the action
```
ouranos: Use [YARD](https://yardoc.org/) format for documentation

eg:
```ruby
# @param user [MnoEnterprise::User] the user performing the action
# @param updated_user [MnoEnterprise::User, MnoEnterprise::OrgaInvite] the user to update
```

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -188,6 +211,17 @@ def user_params
       params.require(:user).permit(:email, :name, :surname, :phone)
     end
 
+    def requested_user_in_org(organization)
+      # If a user has updated their email address, but has not clicked on the confirmation link yet,
+      # we won't find them from their email, so we look from their id first.
+      user = organization.users.find { |u| u.id == params.require(:member)[:id] }
+      user ||= begin
+        email = params.require(:member).require(:email)
+        organization.users.find { |u| u.email == email } ||
```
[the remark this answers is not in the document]
ouranos: Also could you pass the entity and id as params this way you can simplify the controller logic a bit?

PR #578, opened 2017-11-15
[MNOE-209][MNOE-788] Improve Devise configuration

ouranos:
Improve devise default configuration
* Enable account lockout
* Expose devise configuration to the frontend (for password length check)
* Fix password expiry
* Fixed session timeout


ouranos:
@BrunoChauvet could you review the last commit? Thanks

PR #579, opened 2017-11-15
Add impac dashboard feature flag to frontend config settings

adamaziz15:
@ouranos Leaving comments in JIRA, please review :)


ouranos:
See maestrano/mno-enterprise-angular#382

PR #581, opened 2017-11-17
[MNOE-249] Fix Devise::FailureApp routes

ouranos:
The FailureApp was redirecting wihout the engine prefix (/mnoe/) which
was causing 404

### DIFF
--- a/api/config/initializers/devise_patch.rb
+++ b/api/config/initializers/devise_patch.rb
@@ -1,0 +1,39 @@
+require "action_controller/metal"
+
+module Devise
+  # Failure application that will be called every time :warden is thrown from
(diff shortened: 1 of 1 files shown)


api/config/initializers/devise_patch.rb
```diff
@@ -0,0 +1,40 @@
...
+    # Monkey patching to unset opts[:script_name]
+    # See https://github.com/plataformatec/devise/issues/4127
+    def scope_url
+      opts  = {}
+      route = route(scope)
+      opts[:format] = request_format unless skip_format?
+
+      config = Rails.application.config
+
+
```
hedudelgado: double empty line

ouranos: Done

PR #584, opened 2017-11-21
[MNOE-799] Unable to create new customer from Admin Panel

rheasunshine:
provision_app_instances! missing bang.. bug & specs fixed
@ouranos (unable to add a reviewer)


ouranos:
@x4d3 FYI

PR #588, opened 2017-11-23
Impac! permissions

cesar-tonnoir:
- [x] mention dependency to mnohub: https://github.com/maestrano/maestrano-hub/pull/818

- Modify `current_user` so ACL is attached to each organization
- Can pass the impersonator id to get the **impersonator's** access rights instead of the user's ones
- Add CanCan abilities to restrict the actions on Impac! controllers

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
@@ -28,15 +28,10 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::DashboardsControlle
   # POST /mnoe/jpi/v1/impac/dashboards
   #   -> POST /api/mnoe/v1/users/1/dashboards
   def create
-    # TODO: dashboards.build breaks as dashboard.organization_ids returns nil, instead of an
-    #       empty array. (see MnoEnterprise::Impac::Dashboard #organizations)
-    # @dashboard = dashboards.build(dashboard_create_params)
-    # TODO: enable authorization
-    # authorize! :manage_dashboard, @dashboard
-    # if @dashboard.save
+    authorize! :create_impac_dashboards, dashboards.build(dashboard_create_params)
+
     if @dashboard = dashboards.create(dashboard_create_params)
       MnoEnterprise::EventLogger.info('dashboard_create', current_user.id, 'Dashboard Creation', @dashboard)
-
       render 'show'
     else
       render_bad_request('create dashboard', @dashboard.errors)
@@ -47,9 +42,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::DashboardsControlle
   #   -> PUT /api/mnoe/v1/dashboards/1
   def update
     return render_not_found('dashboard') unless dashboard
-
-    # TODO: enable authorization
-    # authorize! :manage_dashboard, dashboard
+    authorize! :update_impac_dashboards, dashboard
 
     if dashboard.update(dashboard_update_params)
       render 'show'
@@ -62,9 +55,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::DashboardsControlle
   #   -> DELETE /api/mnoe/v1/dashboards/1
   def destroy
     return render_not_found('dashboard') unless dashboard
-
-    # TODO: enable authorization
-    # authorize! :manage_dashboard, dashboard
+    authorize! :destroy_impac_dashboards, dashboard
 
     if dashboard.destroy
       MnoEnterprise::EventLogger.info('dashboard_delete', current_user.id, 'Dashboard Deletion', dashboard)
@@ -81,6 +72,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Impac::DashboardsControlle
   # POST mnoe/jpi/v1/impac/dashboards/1/copy
   def copy
     return render_not_found('template') unless template
+    authorize! :create_impac_dashboards, template
 
     # Owner is the current user by default, can be overriden to something else (eg: current organization)
     @dashboard = template.copy(current_user, dashboard_params[:name], dashboard_params[:organization_ids])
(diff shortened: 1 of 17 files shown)


api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
```diff
@@ -28,15 +28,10 @@ def show
   # POST /mnoe/jpi/v1/impac/dashboards
   #   -> POST /api/mnoe/v1/users/1/dashboards
   def create
-    # TODO: dashboards.build breaks as dashboard.organization_ids returns nil, instead of an
-    #       empty array. (see MnoEnterprise::Impac::Dashboard #organizations)
-    # @dashboard = dashboards.build(dashboard_create_params)
-    # TODO: enable authorization
-    # authorize! :manage_dashboard, @dashboard
-    # if @dashboard.save
+    authorize! :create_impac_dashboards, dashboards.build(dashboard_create_params)
```
ouranos: we're doing a double instantiation here.
Could we have?
```ruby
@dashboard = dashboards.build(dashboard_create_params)
authorize! :create_impac_dashboards, @dashboard 
if @dashboard.save
  ...
end
```
Or is it not working as the old commented code is saying?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/kpis_controller.rb
```diff
@@ -49,16 +49,15 @@ def index
...
+  # TODO: nest alert in as a param, with the current user as a recipient.
   def create
     if params[:kpi][:widget_id].present?
       return render_not_found('widget') if widget.blank?
-      authorize! :manage_widget, widget
     else
       return render_not_found('dashboard') if dashboard.blank?
-      authorize! :manage_dashboard, dashboard
     end  
+    authorize! :create_impac_kpis, kpi_parent.kpis.build(kpi_create_params)
```
ouranos: Same here, double instantiation

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/widgets_controller.rb
```diff
@@ -24,6 +24,8 @@ def index
   #  -> POST /api/mnoe/v1/dashboards/:id/widgets
   def create
     if widgets
+      authorize! :create_impac_widgets, widgets.build(widget_create_params)
```
ouranos: Double instantiation

core/app/models/mno_enterprise/impac/dashboard.rb
```diff
@@ -22,13 +22,13 @@ def full_name
     end
 
     # Return all the organizations linked to this dashboard and to which the user has access
-    # If the dashboard is a template, return all the current user's organization
+    # If the dashboard is a template, return all the current user's organizations
     def organizations(org_list = nil)
       if org_list
         return org_list if dashboard_type == 'template'
-        org_list.to_a.select { |e| self.organization_ids.include?(e.uid) }
+        org_list.to_a.select { |e| organization_ids.include?(e.uid) || organization_ids.include?(e.id) }
```
ouranos: Isn't in only `uids` (based on the `where` clause below)?

core/lib/mno_enterprise/concerns/models/ability.rb
```diff
@@ -66,86 +66,69 @@ def initialize(user)
...
-  def impac_abilities(user)
-    can :manage_impac, MnoEnterprise::Impac::Dashboard do |dhb|
-      dhb.organizations.any? && dhb.organizations.all? do |org|
-        !!user.role(org) && ['Super Admin', 'Admin'].include?(user.role(org))
-      end
+  # Enables / disables Impac! Angular capabilities
+  def impac_abilities(orgs_with_acl)
+    can :create_impac_dashboards, MnoEnterprise::Impac::Dashboard do |d|
+      orgs = d.organizations(orgs_with_acl)
+      orgs.present? && orgs.all? { |org| org.acl.dig(:related, :dashboards, :create) }
```
ouranos: `dig` not compatible with ruby < 2.3

core/lib/mno_enterprise/concerns/models/ability.rb
```diff
@@ -66,86 +66,69 @@ def initialize(user)
     end
 
     #===================================================
-    # Impac
+    # Admin abilities
     #===================================================
-    impac_abilities(user)
+    admin_abilities(user)
```
ouranos: Is the order change intentional?

admin_abilities were called last as there supposed to have precedence over everything:

> An ability rule will override a previous one
See: https://github.com/CanCanCommunity/cancancan/wiki/Ability-Precedence

I guess it doesn't matter in this case as the Impac! ACL is coming from MnoHub anyway?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/impac/dashboards_controller.rb
cesar-tonnoir: Indeed it doesn't work as `build + save` does not seem to behave the same as `create`: some parameters (as the owner_id for example) are missing when the `save` Her-built request is sent to mno-hub, as a result, the record is not persisted correctly.

I've tried several things but always run into the same issues so I think we'd better let this code unchanged for now.

core/app/models/mno_enterprise/impac/dashboard.rb
cesar-tonnoir: no: ids are passed when the request is a `create` - this organizations uids/ids thing on the impac components are super confusing and should be refactored accross all the applications

core/lib/mno_enterprise/concerns/models/ability.rb
cesar-tonnoir: it wasn't intentional - reverting

core/app/models/mno_enterprise/impac/dashboard.rb
```diff
@@ -44,10 +44,7 @@ def filtered_widgets_templates
     end
 
     def to_audit_event
-      {
-        name: name,
-        organization_id: (owner_type == 'Organization') ? owner_id : nil
-      }
+      { name: name }
```
ouranos: @cesar-tonnoir didn't spot this one during code review. Why was this removed? I think that'd introduce a regression (see MNOE-861)

cesar-tonnoir: @ouranos I think this is because after this change, owner_type should never be `Organization` again, so `organization_id` would always be nil - I figured out it would be more relevant to simply delete the field?...

PR #589, opened 2017-11-24
[MNOE-809] Add a feature flag for Product Markup

ouranos:
Add feature flag and missing routing specs


ouranos:
Rebased on top of #590 to fix the build

PR #590, opened 2017-11-27
JsonApiClient fixes for 1.5.3

ouranos:
* Bump min version to 1.5.3
* Default pagination params changed in 1.5.3. See https://github.com/chingor13/json_api_client/pull/267
* Remove JsonApiClient monkey patch. See https://github.com/chingor13/json_api_client/pull/275


ouranos:
@x4d3 FYI. I've removed your patch. Can you double check I haven't missed anything?

PR #592, opened 2017-11-29
[MNOE-812] Expose system identity on jpi

BrunoChauvet:


core/app/models/mno_enterprise/system_identity.rb
```diff
@@ -0,0 +1,14 @@
+module MnoEnterprise
+  class SystemIdentity < BaseResource
+    def self.table_name
+      'system_identity'
+    end
+
+    def to_audit_event
```
ouranos: Not sure you need this as you won't log any event on this?

BrunoChauvet: true

PR #595, opened 2017-12-04
[MNOE-819] Update error handling when inviting user to organization

adamaziz15:
@ouranos Ready for review. Leaving comments in JIRA.


ouranos:
Can you rebase and open the PR against `3.3`? Thanks

adamaziz15:
@ouranos Updated to branch 3.3

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -73,14 +73,19 @@ def invite_member
       user = MnoEnterprise::User.find_by(email: user_params[:email]) || create_unconfirmed_user(user_params)
 
       # Create the invitation
-      invite = @organization.org_invites.create(
+      invite = MnoEnterprise::OrgInvite.create(
```
ouranos: Why this change?

api/spec/controllers/mno_enterprise/jpi/v1/admin/organizations_controller_spec.rb
```diff
@@ -157,7 +157,7 @@ def partial_hash_for_arrears(arrear)
         # Track the api call
         @api_call = false
         stub = -> { @api_call = true; from_api(org_invite) }
-        api_stub_for(post: "/organizations/#{organization.id}/org_invites", response: stub)
+        api_stub_for(post: "/org_invites", response: stub)
```
ouranos: We need a spec for when the invitation is not saved.
I think `persisted?` is just checking the `id` so you can have a stub without id for this case.

PR #596, opened 2017-12-05
[OPAL-261] Fix upload issue

x4d3:
- sub_tenant was not loaded, triggering a 404 when api/app/views/mno_enterprise/jpi/v1/admin/users/_user.json.jbuilder was called

### DIFF
--- a/api/lib/mno_enterprise/csv_importer.rb
+++ b/api/lib/mno_enterprise/csv_importer.rb
@@ -60,13 +60,12 @@ module MnoEnterprise
         # Create or Update User
         user = MnoEnterprise::User.where(email: row['email']).first
         event_type = if user
-                       :added
+                       :updated
                      else
                        user = MnoEnterprise::User.new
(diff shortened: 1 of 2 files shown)


api/lib/mno_enterprise/csv_importer.rb
```diff
@@ -59,19 +59,22 @@ def self.process(file_path)
...
-        if user
-          report[:users][:updated] << user
-        else
-          user = MnoEnterprise::User.new
-          report[:users][:added] << user
-          user.password = Devise.friendly_token
-          user.email = row['email']
-        end
+        event_type = if user
+                       :added
```
ouranos: Shouldn't it be `:updated` in this case.
I'm a bit confused with this part of the code.

x4d3: Well spotted.

x4d3:
@ouranos I've fixed + added more spec

PR #602, opened 2017-12-08
[MNOE-833] - Fix wrong entity sent back on user creation

fgourichon:


fgourichon:
I think it should be ported on 4.0 @ouranos

ouranos:
Thanks. 
Yes I'll merge  `3.3` into `4.0` later

PR #603, opened 2017-12-12
[MNOE-776] Fix issues with invitation management

manu-d:
- For the invite_member, organisation.orga_relations were not up to date with the new orga relation created from the orga invite. Then, when returning the json, this line was nil: https://github.com/manu-d/mno-enterprise/blob/4.0/api/app/views/mno_enterprise/jpi/v1/admin/base_resource/_member.json.jbuilder#L4, so we couldn't get the value on the frontend


ouranos:
Build is broken

PR #610, opened 2017-12-14
[MNOE-732] Support provisioning via /product_instance endpoint

enizor:


ouranos:
This has been integrated in the `feature/json-api-proxy` branch

PR #612, opened 2017-12-18
[MNOE-110] Handle MnoHub Errors

ouranos:
Better handle when maestrano-hub is down or slow to respond.

We rescue Faraday errors and return an appropriate error code:
* 503 Service Unavailable when maestrano-hub is down
* 429 Too Many Request on connection timeout


ouranos:
Fixed locales

PR #614, opened 2017-12-18
[MNOE-807] Set max heights for images

manu-d:
html max-height and max-width are not recognised by some mail clients like Outlook, so we calculate them ourselves

### DIFF
--- a/core/app/helpers/mno_enterprise/image_helper.rb
+++ b/core/app/helpers/mno_enterprise/image_helper.rb
@@ -1,8 +1,10 @@
+require 'fastimage'
+
 module MnoEnterprise
   module ImageHelper
-    
+
     IMAGES_LOCATION = "/app/assets/images/mno_enterprise/"
-    
+
     # Helper method to easily access and select the images
     # If full_path is true returns filename appended to the path
     # If full_path is false returns filename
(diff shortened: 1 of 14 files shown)


core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -32,5 +34,27 @@ def engine_image_path(file_name)
     def fit_image
       'max-width: 150px; max-height: 150px;'
     end
+
+    def main_logo_white_bg_dimensions
+      return FastImage.size(main_logo_white_bg_path(true))
```
ouranos: No need for return

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -32,5 +34,27 @@ def engine_image_path(file_name)
...
+    end
+
+    # Returns dimensions for the main_logo_white_bg, with width < 150px && height < 150px
+    # while keeping the same aspect ratio
+    def main_logo_white_bg_size_to_fit
+      dimensions = main_logo_white_bg_dimensions
+      original_width = dimensions[0]
+      original_height = dimensions[1]
```
ouranos: `original_width, original_height = dimensions`

core/app/helpers/mno_enterprise/image_helper.rb
```diff
@@ -32,5 +34,27 @@ def engine_image_path(file_name)
...
+      return dimensions if original_width < 150 && original_height < 150
+
+      if original_width > original_height
+        width = [original_width, 150].min
+        height = width * original_height / original_width
+      else
+        height = [original_height, 150].min
+        width = height * original_width / original_height
+      end
+      [width, height]
```
ouranos: Would it be better to return a hash? so you can use the return value straight away?

PR #615, opened 2017-12-20
[MNOE-744] Better authorisation error reporting

ouranos:
This enables us to differentiate between unauthenticated users and authenticated users which are not admin. We'll be able to handle them differently in the frontend (redirect to sign_in or customer view).

**Dependency:**
This requires a change in the admin panel to handle 403 errors.  Do not merge.
Fixed in maestrano/mnoe-admin-panel#189


ouranos:
@x4d3 have a look at the first commit for the subtenant specs and see MNOE-856

PR #618, opened 2017-12-26
[MNOE-857] Allow current user to be assigned to the organization created from Admin menu

MAhsenArif:


MAhsenArif:
Closing as fix needs to be done on master branch, which has different structure.

PR #621, opened 2018-01-01
[MNOE-427] Refresh user cache after a member role is updated in an organization

MAhsenArif:


ouranos:
Closing as this wouldn't fix this specific problem as discussed in JIRA

PR #624, opened 2018-01-04
[MNOE-821] Expose Tenant#metadata

enizor:
It allows to set the Admin Panel Settings/apps screen Read-Only when metadata.tenant_management is not "tenant"


enizor:
Fixed it, will squash if the rest seems good

enizor:
Squashed

PR #625, opened 2018-01-04
[MNOE-865] Send patch request to correct MnoHub endpoints for reviews

manu-d:
On MnoHub side, json_api_resource needs the type of the payload to be the same as the type in the controller path, i.e 
<pre><json>
{  
   "data":{  
      "id":"1",
      "type":"<b>comments</b>",
      "attributes":{  
         "description":"Great app",
         "user_id":"1"
      }
   }
}
</pre>
needs to be sent to /api/mnoe/v2/**comments**/1

### DIFF
--- a/core/app/models/mno_enterprise/review.rb
+++ b/core/app/models/mno_enterprise/review.rb
@@ -3,5 +3,18 @@ module MnoEnterprise
     property :created_at, type: :time
     property :updated_at, type: :time
     property :user_id, type: :string
+
+    # Since Review is the parent metaclass for Comment, Feedback, Question & Answer,
+    # we need to send the request to the appropriate endpoint
+    def self.path(params = nil)
+      @type_param = params[:type] if params
+      res = super
+      @type_param = nil
+      res
+    end
+
+    def self.resource_path
+      @type_param || super
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


core/app/models/mno_enterprise/review.rb
```diff
@@ -3,5 +3,19 @@ class Review < BaseResource
     property :created_at, type: :time
     property :updated_at, type: :time
     property :user_id, type: :string
+
+    # Since Review can be a Comment or a Feedback, we need to send
```
ouranos: or a Question :)
Maybe put something like:
```
# Since Review is the parent metaclass for Comment, Feedback, Question & Answer,
# we need to send the request to the appropriate endpoint
```

core/app/models/mno_enterprise/review.rb
```diff
@@ -3,5 +3,19 @@ class Review < BaseResource
...
+    def self.path(params = nil)
+      @type_param = params[:type] if params
+      res = super
+      @type_param = nil
+      res
+    end
+
+    def self.resource_path
+      return @type_param if @type_param
+      super
```
ouranos: `@type_param || super`

manu-d:
@ouranos Can you approve for merge?

PR #626, opened 2018-01-05
[MNOE-644] Update controllers and views to use pricing plans as product pricings

adamaziz15:
@ouranos

### DIFF
--- a/api/app/views/mno_enterprise/jpi/v1/admin/products/_product.json.jbuilder
+++ b/api/app/views/mno_enterprise/jpi/v1/admin/products/_product.json.jbuilder
@@ -14,7 +14,13 @@ json.assets_attributes do
 end
 
 json.product_pricings do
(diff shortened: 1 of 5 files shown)


BrunoChauvet:
It makes sense to expose both on the MnoHub API but I believe the front-end needs only the prices including the user-related markup/discount

adamaziz15:
@ouranos The frontend is using `product_pricings` at the moment, would need to update it to use `pricing_plans`. @BrunoChauvet The markup/discounts are applied to both `product_pricings` as well as `pricing_plans`. The only difference being that `pricing_plans` would be localised pricings.

adamaziz15:
Updated so that `pricing_plans` will be returned to mnoe-ng, but `product_pricings` will still be returned to the admin panel since all pricings should be returned there, not just localised ones.

PR #629, opened 2018-01-08
[MNOE-470] Currency Selection (Customer / Tenant)

MAhsenArif:


MAhsenArif:
@manu-d Updated locales

PR #634, opened 2018-01-16
[MNOE-806] Ability to skip depedencies for products

manu-d:
- Ability to fetch only a set of fields in base_resource
- Ability to skip fetching all dependencies for products

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/products_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/products_controller.rb
@@ -16,7 +16,8 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Admin::ProductsController
       JSON.parse(params[:terms]).map { |t| @products = @products | MnoEnterprise::ProductMarkup.includes(DEPENDENCIES).where(Hash[*t]).where(active: true) }
       response.headers['X-Total-Count'] = @products.count
     else
-      query = MnoEnterprise::Product.apply_query_params(params).where(active: true).includes(DEPENDENCIES)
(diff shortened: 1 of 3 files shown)


api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/products_controller.rb
```diff
@@ -16,7 +16,8 @@ def index
       JSON.parse(params[:terms]).map { |t| @products = @products | MnoEnterprise::ProductMarkup.includes(DEPENDENCIES).where(Hash[*t]).where(active: true) }
       response.headers['X-Total-Count'] = @products.count
     else
-      query = MnoEnterprise::Product.apply_query_params(params).where(active: true).includes(DEPENDENCIES)
+      query = MnoEnterprise::Product.apply_query_params(params).where(active: true)
+      query = query.includes(DEPENDENCIES) unless params[:skip_dependencies]
```
BrunoChauvet: @ouranos @manu-d 
Shouldn't it be the front-end that would specify the included fields?
`query.includes(params[:includes])` so we can tune queries from the front-ends

manu-d: It would be even better to put it in the method apply_query_params (https://github.com/maestrano/mno-enterprise/blob/4.0/core/app/models/mno_enterprise/base_resource.rb#L28), but maybe as part as a bigger refactoring, since almost all the controllers act like this at the moment

BrunoChauvet: agreed

PR #532, opened 2017-10-13
[MNOE-689] API V2 - SubTenant replace mass assignment of relation ship by diff

x4d3:
- remove clients and account_managers dependency from sub_tenant
- remove clients dependeny from user
- introduce update_clients and update_account_managers endpoint
- replace user.mnoe_sub_tenant_id by sub_tenant_id
- add optional belong_to_subtenant and belong_to_account_manager boolean parameter depending on the sub_tenant_id and account_manager parameter

PR #640, opened 2018-01-25
[MNOE-857] Allowed organization to be added to staff divisions

MAhsenArif:


api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -141,5 +142,19 @@ def update_app_list
         existing_apps.reload
       end
     end
+
+    def update_sub_tenants
+      return if params[:organization].key?(:sub_tenant_ids).blank?
```
ouranos: `key?` returns `true` or `false` so no need for `blank?` here

api/app/controllers/mno_enterprise/jpi/v1/admin/organizations_controller.rb
```diff
@@ -141,5 +142,19 @@ def update_app_list
...
+
+      sub_tenants = MnoEnterprise::SubTenant.where({'id.in' => params[:organization][:sub_tenant_ids]})
+
+      sub_tenants.to_a.each do |sub_tnt|
```
ouranos: This is not really scalable if we have a huge list for `client_ids`.
See MNOE-869 and the fix added in #532

PR #644, opened 2018-02-01
[MNOE-406] Added logic to load delisted apps if already in dashboard

MAhsenArif:


api/app/views/mno_enterprise/jpi/v1/app_instances/_resource.json.jbuilder
```diff
@@ -23,6 +23,7 @@ end
 #   json.connector_version app_instance.oauth_keys[:version]
 # end
 
+app_instance.without_tenant = true if params[:data] && params[:data] == 'unscoped'
```
ouranos: You could use `unscoped` as the params and it'd be:
`app_instance.without_tenant = params[:unscoped].present?`

api/spec/controllers/mno_enterprise/jpi/v1/app_instances_controller_spec.rb
```diff
@@ -47,6 +51,15 @@ module MnoEnterprise
         end
       end
 
+      context 'with unscoped data' do
+        before { allow_any_instance_of(MnoEnterprise::AppInstance).to receive(:without_tenant).and_return(true) }
+
+        it 'retrived the app instance with the app' do
```
ouranos: `retrieved`

api/spec/controllers/mno_enterprise/jpi/v1/admin/organizations_controller_spec.rb
```diff
@@ -43,11 +43,13 @@ def partial_hash_for_arrears(arrear)
     let(:org_invite) { build(:org_invite, organization: organization) }
     let(:app_instance) { build(:app_instance, organization: organization) }
     let(:credit_card) { build(:credit_card, organization: organization) }
+    let(:app) { build(:app)}
 
     before do
       organizations = [organization]
       allow(organizations).to receive(:loaded?).and_return(true)
       allow_any_instance_of(MnoEnterprise::User).to receive(:organizations).and_return(organizations)
+      allow_any_instance_of(MnoEnterprise::AppInstance).to receive(:app).and_return(app)
```
ouranos: Doesn't that mean we have a N+1 here?
* 1 Org with N App Instances
* Fetch N apps?

api/spec/controllers/mno_enterprise/jpi/v1/app_instances_controller_spec.rb
```diff
@@ -47,6 +51,15 @@ module MnoEnterprise
         end
       end
 
+      context 'with unscoped data' do
+        before { allow_any_instance_of(MnoEnterprise::AppInstance).to receive(:without_tenant).and_return(true) }
+
+        it 'retrieved the app instance with the app' do
```
ouranos: I don't think this spec is testing much since you're stubbing `AppInstance#app`

It should be something like this (comments inlined):
```ruby
# remove the MnoEnterprise::AppInstance#without_tenant and MnoEnterprise::AppInstance#app stubs
context 'with unscoped data' do
  subject { get :index, organization_id: organization.id, timestamp: timestamp, unscoped: true }

  it 'retrieves the unscoped app' do
    # Test that we do the correct request to mnohub
    expect(MnoEnterprise::App).to receive(:find).with(app.id, {unscoped: true).and_return(app)
    subject
    # This test is pretty useless as we're testing something stubbed but let's keep it
    expect(app_instance.app).to eq(app)
  end
end
```

PR #646, opened 2018-02-05
[MNOE-908] Added Fields to Organization Info Page

iseessel:
This pull request is directly tied to https://github.com/maestrano/mnoe-admin-panel/pull/212


api/app/views/mno_enterprise/jpi/v1/admin/organizations/_credit_card.json.jbuilder
```diff
@@ -1,6 +1,7 @@
 json.credit_card do
-  if credit_card.try(:id)
+  if credit_card
     json.presence true
+    json.extract! credit_card, :updated_at
```
ouranos: In this case, I think you can use `json.updated_at credit_card.updated_at` which is more explicit
`extract!` is good when extracting multiple fields

PR #648, opened 2018-02-07
[MNOE-909]Removed trackable for impersonation requests

iseessel:
Devise makes it easy to skip trackable. Just need to set request.env["devise.skip_trackable"] = true, BEFORE authentication happens. Tested locally and works.

It is appropriate to put it in the impersonate_controller, rather than the general auth-controller because we specifically want to do this on impersonating. 

If we wanted to skip trackable other times, we could consider moving it into `sessions_controller.rb` and having an `if params[:skip_trackable]`.

### DIFF
--- a/api/app/controllers/mno_enterprise/impersonate_controller.rb
+++ b/api/app/controllers/mno_enterprise/impersonate_controller.rb
@@ -2,6 +2,7 @@ module MnoEnterprise
   class ImpersonateController < ApplicationController
     include MnoEnterprise::ImpersonateHelper
 
+    before_filter :skip_trackable, only: [:create]
     before_filter :authenticate_user!, except: [:destroy]
     before_filter :current_user_must_be_admin!, except: [:destroy]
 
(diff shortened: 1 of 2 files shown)


adamaziz15 (dismissed):
LGTM 👍

api/app/controllers/mno_enterprise/impersonate_controller.rb
```diff
@@ -39,6 +40,10 @@ def destroy
 
     private
 
+    def skip_trackable
```
ouranos: This part is not needed, the function is already defined in the `ApplicationController`

ouranos (changes requested):

api/app/controllers/mno_enterprise/impersonate_controller.rb
iseessel: Ah I see. 

I'm a little confused as to why this is in the application_controller. Trackables are only updated when the user signs in, correct? I see several `before_filter :authenticate_user!`, does it have to do with that?

ouranos: Sorry this was added in #578 and hasn't made to the `4.0` branch yet.
The method is in the `ApplicationController` so it can be called from any controller that might requires it.

Added more details in JIRA

PR #649, opened 2018-02-13
Admin::OrganizationsController: Moved logic to concern

MAhsenArif:
`OrganizationsController#freeze` has been renamed to `OrganizationsController#freeze_account` since `freeze` is being recognized as a keyword after extracting controller logic to concern


api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/organizations_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+    elsif user.is_a?(MnoEnterprise::OrgaInvite)
+      user.decline!
+    end
+
+    render 'members'
+  end
+
+  # PUT /mnoe/jpi/v1/admin/organizations/1/freeze
+  # Note: We need to rename this method after moving to concern,
+  #       since freeze is recognized as a keyword
```
manu-d: Can you remove this comment since you renamed the method?

manu-d: This should be a comment in the pull request, not in the code

PR #651, opened 2018-02-19
[MNOE-926] Fix Dashboards backward compatibility

ouranos:
Since ids are Integer on APIv1 and String on APIv2, this causes the
Dashboard#sorted_widgets to create on APIV2 for dashboards created and
sorted on APIv1.

### DIFF
--- a/core/app/models/mno_enterprise/dashboard.rb
+++ b/core/app/models/mno_enterprise/dashboard.rb
@@ -53,7 +53,7 @@ module MnoEnterprise
     end
 
     def sorted_widgets
-      ids = self.widgets_order | self.widgets.map(&:id)
+      ids = self.widgets_order.map(&:to_s) | self.widgets.map(&:id)
       widgets_per_ids = self.widgets.each_with_object({}) do |w, hash|
         hash[w.id] = w
       end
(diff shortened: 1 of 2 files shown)


core/app/models/mno_enterprise/dashboard.rb
```diff
@@ -53,7 +53,7 @@ def filtered_widgets_templates
     end
 
     def sorted_widgets
-      ids = self.widgets_order | self.widgets.map(&:id)
+      ids = self.widgets_order.map(&:to_s) | self.widgets.map(&:id)
```
cesar-tonnoir: Will this still work with API v1 ? Won't `self.widgets.map(&:id)` return integers and `self.widgets_order.map(&:to_s)` strings?

core/spec/models/mno_enterprise/dashboard_spec.rb
```diff
@@ -11,7 +11,30 @@ module MnoEnterprise
     end
 
     describe '#sorted_widgets' do
-      it 'is pending'
+      let(:w1) { build(:impac_widget) }
```
cesar-tonnoir: thanks

PR #653, opened 2018-02-22
[KAPP-114] Added new action to accept TOS for a user

MAhsenArif:


ouranos:
@MAhsenArif  specs are red

MAhsenArif:
@ouranos Added test-cases

PR #632, opened 2018-01-15
[MNOE-872] Expose session timeout config

enizor:
Allows the frontend to warn the user when he will get logged off due to inactivity

PR #654, opened 2018-02-22
[MNOE-872] Expose session timeout config

MAhsenArif:
Allows the frontend to warn the user when he will get logged off due to inactivity

Continued https://github.com/maestrano/mno-enterprise/pull/632 in this PR

### DIFF
--- a/api/app/views/mno_enterprise/auth/sessions/new.html.haml
+++ b/api/app/views/mno_enterprise/auth/sessions/new.html.haml
@@ -3,7 +3,7 @@
     .row
       - if params[:session_timeout].present?
         .alert.alert-warning.text-center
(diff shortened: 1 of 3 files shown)


api/app/views/mno_enterprise/auth/sessions/new.html.haml
```diff
@@ -1,6 +1,9 @@
 .registration
   .container
     .row
+      - if params[:session_timeout].present?
+        .alert.alert-warning.text-center
+          You are logged off due to inactivity. Please log in again
```
manu-d: You should add a translation in this file:
https://github.com/maestrano/mno-enterprise/blob/master/core/config/locales/views/auth/sessions/en.yml

PR #657, opened 2018-02-27
[MNOE-916] Fix file path for sample csv download

paulstuartparker:


api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/organizations_controller.rb
```diff
@@ -194,7 +194,7 @@ def unfreeze
 
   # GET /mnoe/jpi/v1/admin/organizations/download_batch_example
   def download_batch_example
-    path = File.join(File.dirname(File.expand_path(__FILE__)), '../../../../../assets/batch-example.csv')
+    path = File.join(File.dirname(File.expand_path(__FILE__)), '../../../../../../../app/assets/batch-example.csv')
```
ouranos: Huh guys, could we use this instead:
```ruby
path = MnoEnterprise::Api::Engine.root.join('app/assets/batch-example.csv')
```

api/spec/controllers/mno_enterprise/jpi/v1/admin/organizations_controller_spec.rb
```diff
@@ -202,11 +202,24 @@ module MnoEnterprise
       it { expect(response).to be_success }
     end
 
+    describe 'GET #download_batch_example' do
+
+      subject { get :download_batch_example, file: file }
+
+      context 'successful download' do
+        let(:path) { MnoEnterprise::Api::Engine.root.join('spec/fixtures/batch-download-example.csv') }
```
ouranos: Do you need to introduce an extra file here?
Why not reuse `MnoEnterprise::Api::Engine.root.join('app/assets/batch-example.csv')`?

PR #663, opened 2018-03-05
[MNOE-845] Internationalization default webstore language fix

MAhsenArif:


core/lib/mno_enterprise/concerns/controllers/i18n.rb
```diff
@@ -3,11 +3,11 @@ module MnoEnterprise::Concerns::Controllers::I18n
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_locale, if: "MnoEnterprise.i18n_enabled"
+    before_action :set_locale
   end
 
   def set_locale
-    I18n.locale =  if params[:locale] && I18n.available_locales.include?(params[:locale].to_sym)
+    I18n.locale =  if params[:locale] && I18n.available_locales.include?(params[:locale].to_sym) &&MnoEnterprise.i18n_enabled
```
ouranos: Missing a space after `&&`

api/app/controllers/mno_enterprise/config_controller.rb
```diff
@@ -46,5 +47,9 @@ def available_locales
         }
       end
     end
+
+    def set_default_locale
```
ouranos: Would it be better do to this in [TenantConfig](https://github.com/maestrano/mno-enterprise/blob/4.0/core/app/models/mno_enterprise/tenant_config.rb) where we are reloading the settings?

MAhsenArif: Discussed on Slack. TenantConfig does not have the available locales list on app load and cannot set default_locale.

PR #664, opened 2018-03-06
[MNOE-941]Fix org admin invoices

iseessel:
1) Change jbuilder to reflect the invoice path for org admins vs. tenant-wide admins 
2) Pass in the organization to the authorize! method, instead of an enumerator class of JSON API (#find for JSON API doesn't return a single record like in Rails).
3) Move the endpoint for organization invoices into the api engine, under jpi/v1/invoices namespace. 

Tested locally.


iseessel:
@adamaziz15  @ouranos I'm investigating travis build.

PR #666, opened 2018-03-07
[MNOE-928] Add specs

ouranos:
Forgot to commit some specs

### DIFF
--- a/api/spec/controllers/mno_enterprise/pages_controller_spec.rb
+++ b/api/spec/controllers/mno_enterprise/pages_controller_spec.rb
@@ -45,15 +45,45 @@ module MnoEnterprise
     end
 
     describe 'GET #loading' do
-      before do
-        stub_api_v2(:get, '/app_instances', [app_instance], [:app], {filter:{uid: app_instance.uid}, page:{number: 1, size: 1}})
-      end
-
-      # before { sign_in user }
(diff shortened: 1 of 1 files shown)


api/spec/controllers/mno_enterprise/pages_controller_spec.rb
```diff
@@ -45,15 +45,45 @@ module MnoEnterprise
...
+              'errors' => [],
+              'logo' => app_instance.app.logo
+            }
+            expect(JSON.parse(subject.body)).to eq(expected_hash)
+          end
+        end
+
+        context 'when the application is not found' do
+          before do
+            stub_api_v2(:get, '/app_instances', [], [:app], {filter: {uid: app_instance.uid}, page: {number: 1, size: 1}})
```
BrunoChauvet: Should it be `page: {number: 1, size: 0}` ? not a big deal if this works

ouranos: This hash reflect the client params. So
```
MnoEnterprise::AppInstance.where(uid: params[:id]).includes(:app).first
```
translates to `page: {number: 1, size: 1}`

BrunoChauvet: ok my bad

PR #672, opened 2018-03-23
[MNOE-984] CSV Empty User Address

paulstuartparker:
Ensures that the User will be associated with an address.  Currently the address is only associated with the Organization on the CSV import.


ouranos:
Closing as per discussion in JIRA

PR #673, opened 2018-03-26
[MNOE-978] Refactor pagination

iseessel:


adamaziz15 (approved):

api/app/views/mno_enterprise/jpi/v1/admin/organizations/_organization.json.jbuilder
```diff
@@ -1,2 +1,2 @@
 json.extract! organization, :id, :name, :uid, :soa_enabled, :created_at, :account_frozen, :financial_metrics, :billing_currency, :external_id, :belong_to_sub_tenant, :belong_to_account_manager, :demo_account
-
+json.role @user.role(organization) if @user
```
adamaziz15: Since `@user` is being passed into the partial in the view (...admin/organizations/show.json.jbuilder), can it be referred to as just `user` here or can the `user: @user` be removed from the show view?

api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/organizations_controller.rb
```diff
@@ -26,6 +26,10 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::Admin::OrganizationsContro
   #==================================================================
   # GET /mnoe/jpi/v1/admin/organizations
   def index
+    # When fetching multiple organizations attached to one user, we must fetch the user's role for each organization.
+    where = params["where"]
+    @user = MnoEnterprise::User.find_one(params["where"]["users.id"], :orga_relations) if where && where["users.id"]
```
adamaziz15: Can we refactor this like this maybe? 
```
@user = MnoEnterprise::User.find_one(params["where"]["users.id"], :orga_relations) if params.dig('where', 'users.id')
```
Then the assignment on line 30 can be removed.

adamaziz15 (changes requested):

adamaziz15 (approved):
lgtm

ouranos (dismissed):

PR #677, opened 2018-03-30
[MNOE-996] [MNOE-997] [MNOE-977] - App Management Page

MAhsenArif:


api/app/views/mno_enterprise/jpi/v1/product_instances/_resource.json.jbuilder
```diff
@@ -0,0 +1,31 @@
+json.id product_instance.id
+json.uid product_instance.uid
+json.name product_instance.product.name
+json.status product_instance.status
+json.created_at product_instance.created_at
+
+# Note: Handles tiny_description, getting_started, support_url from product#values
+product_instance.product.values.tap do |value|
+  json.value&.field&.name value&.data
```
BrunoChauvet: I think you should use the attribute `json.value&.field&.nid`

api/lib/mno_enterprise/concerns/controllers/jpi/v1/product_instances_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    @product_instances = MnoEnterprise::ProductInstance.includes(*DEPENDENCIES).where(organization_id: parent_organization.id, 'status.in': statuses, 'fulfilled_only': true).to_a.select do |i|
+      can?(:access_product_instance, i)
+    end
+  end
+
+  # TODO: Copied over structure for AppInstances.
+  #       Needs to be setup for ProductInstances when doing complete migration
+  #       from AppInstance to ProductInstance
+
+  # POST /mnoe/jpi/v1/organization/1/product_instances
```
BrunoChauvet: We should not support the create/destroy actions anyway as products will be managed via subscriptions

core/lib/mno_enterprise/concerns/models/product_instance.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  # TODO: Copied over structure from AppInstance.
+  #       Needs to be filled when doing complete migration from AppInstance to
+  #       ProductInstance
+
+  #==================================================================
+  # Class methods
+  #==================================================================
+  module ClassMethods
```
BrunoChauvet: The provision! and terminate! actins may not be required neither and managed via subscriptions

PR #679, opened 2018-04-03
[MNOE-1020] Pagination broken

iseessel:
@BrunoChauvet @ouranos @adamaziz15 

This was changed in https://github.com/maestrano/mno-enterprise/commit/4046ddf2ca5ed55c5d22accb3fa792af6befb50c and I'm a bit confused as to why; is there a corresponding issue I could look at?  

When using #fetch_all it ignores pagination (expected behavior I assume). 

I'm a bit confused as to the purpose of the fetch_all method, because I did some testing and changing it to '.to_a' returns all the records when not given pagination parameters, but I think I'm missing something.


BrunoChauvet:
My understanding is that there is a default pagination made on the server side when the page details are not specified. If I remember correctly this should return the first 25 records. You should test with more than 25 products and verify that all of them are returned.

iseessel:
@BrunoChauvet Thanks for the input Bruno -- will test.

ouranos:
Indeed, the server will paginate by default, the `fetch_all` will loop through the page to retrieve everything.
This is a workaround as we don't have pagination yet on the marketplace.

ouranos:
Rebased to solve conflicts

iseessel:
@ouranos thanks!

PR #683, opened 2018-04-06
[OPAL-432] Fetch appropriate json schemas

iseessel:


api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/products_controller.rb
```diff
@@ -25,7 +25,11 @@ def index
 
   # GET /mnoe/jpi/v1/admin/products/id
   def show
-    @product = MnoEnterprise::Product.find_one(params[:id], DEPENDENCIES)
+    @product = MnoEnterprise::Product
```
BrunoChauvet: Do we need to deal with the scenario where `editAction` is not specified?

iseessel: @BrunoChauvet So there is guarding on the adapter bluesky for when the edit_action isn't specified -- nothing should break on mnoe when its not specified, but we may want to put a default here as well.

api/app/views/mno_enterprise/jpi/v1/admin/subscriptions/_subscription.json.jbuilder
```diff
@@ -9,6 +9,7 @@ json.available_licenses subscription.available_licenses
 json.external_id subscription.external_id
 json.custom_data subscription.custom_data
 json.provisioning_data subscription.provisioning_data
+json.available_edit_actions subscription.available_edit_actions
```
BrunoChauvet: Before we forget, this has been renamed to `available_actions`
https://github.com/maestrano/maestrano-hub/pull/972/files#diff-c8fbb65db37a91bbb5be88241f81e874R24

api/app/views/mno_enterprise/jpi/v1/subscriptions/_subscription.json.jbuilder
```diff
@@ -9,6 +9,7 @@ json.available_licenses subscription.available_licenses
 json.external_id subscription.external_id
 json.custom_data subscription.custom_data
 json.provisioning_data subscription.provisioning_data
+json.available_edit_actions subscription.available_edit_actions
```
BrunoChauvet: Same comment as above

api/app/views/mno_enterprise/jpi/v1/admin/subscriptions/_subscription.json.jbuilder
iseessel: @BrunoChauvet Oops, yeah I had changed it but not pushed up the changes.

PR #686, opened 2018-04-12
[MNOE-992] AppMetric: Passed organization ids to limit staff users metrics apps

MAhsenArif:


api/app/controllers/mno_enterprise/jpi/v1/admin/app_metrics_controller.rb
```diff
@@ -3,15 +3,22 @@ class Jpi::V1::Admin::AppMetricsController < Jpi::V1::Admin::BaseResourceControl
 
     # GET /mnoe/jpi/v1/admin/app_metrics
     def index
+      org_ids = current_user.organizations.collect(&:id) if current_user.admin_role == 'staff'
```
ouranos: > Prefer `map` over `collect`
https://github.com/bbatsov/ruby-style-guide#map-find-select-reduce-size

ouranos: Btw, this shouldn't be `User#organizations` as this is the list of organization the admin belongs to but `User#clients` which represent the account manager assignments.

PR #687, opened 2018-04-17
[MNOE-970] Create Nex Job to check if restart is done

manu-d:
- When tenant config is updated, create a job on the Nex app to check the config_timestamp.

(Build fails for now. Once PR on nex_client is merged and new version is tagged, will need to update the gem)

### DIFF
--- a/api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
+++ b/api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
@@ -10,9 +10,10 @@ module MnoEnterprise
 
     # PATCH /mnoe/jpi/v1/admin/tenant
     def update
+      timestamp = Time.current.to_i
+      params[:tenant].deep_merge!(frontend_config: { config_timestamp: timestamp })
+
       @tenant = MnoEnterprise::Tenant.show
-      timestamp = Time.now.utc.to_i
-      tenant_params['frontend_config']&.merge!(config_timestamp: timestamp)
       @tenant.update_attributes!(tenant_params)
 
       MnoEnterprise::SystemManager.restart(timestamp)
--- a/api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
+++ b/api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
@@ -11,10 +11,18 @@ module MnoEnterprise
     # PATCH /mnoe/jpi/v1/admin/tenant
     def update
       @tenant = MnoEnterprise::Tenant.show
+      timestamp = Time.now.utc.to_i
+      tenant_params['frontend_config']&.merge!(config_timestamp: timestamp)
       @tenant.update_attributes!(tenant_params)
 
-      MnoEnterprise::SystemManager.restart
+      MnoEnterprise::SystemManager.restart(timestamp)
       render :show
+    end
+
+    # GET /mnoe/jpi/v1/admin/tenant/restart_status
+    def restart_status
+      status = MnoEnterprise::SystemManager.restart_status
+      render json: { status: status }
     end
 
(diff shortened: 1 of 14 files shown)


core/lib/mno_enterprise/platform_adapters/nex_adapter.rb
```diff
@@ -125,7 +152,7 @@ def setup_nex_client
         def nex_app
           @nex_app ||= begin
             setup_nex_client
-            NexClient::App.find(ENV['SELF_NEX_APP_ID']).first
+            NexClient::App.where(name: ENV['SELF_NEX_APP_NAME']).first
```
manu-d: I changed SELF_NEX_APP_ID to SELF_NEX_APP_NAME because filtering on exec_cmds has been setup to filter on executor's name and not id

core/lib/mno_enterprise/platform_adapters/nex_adapter.rb
```diff
@@ -33,15 +33,38 @@ class << self
...
+          cmd = "touch tmp/restart.txt && " \
+            "timestamp=0 && " \
+            "while [ $timestamp -ne #{timestamp} ]; " \
+            "do aux=$(echo `curl -s -X GET http://localhost/mnoe/config.json`) && " \
+            "timestamp=${aux:-0}; done"
+          c = NexClient::ExecCmd.new(
+            name: "check restart timestamp",
+            script: cmd,
+            local: true,
+            parallel: true
```
ouranos: Does that mean all containers restart at once?
Probably not the best as that might cause some downtime while puma restart?
If you put `parallel: false` does that might the job execute sequentially and will wait for all containers to have restarted before exiting? That'd probably be a better way.

api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
```diff
@@ -11,12 +11,20 @@ def show
     # PATCH /mnoe/jpi/v1/admin/tenant
     def update
       @tenant = MnoEnterprise::Tenant.show
+      timestamp = Time.now.utc.to_i
```
ouranos: Prefer to use `Time.current.to_i`

api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
```diff
@@ -11,12 +11,20 @@ def show
...
+      tenant_params['frontend_config']&.merge!(config_timestamp: timestamp)
```
ouranos: If `tenant_params['frontend_config']` is `nil`, the timestamp is not updated so we'll wait forever for the cube to restart.

core/lib/mno_enterprise/platform_adapters/nex_adapter.rb
```diff
@@ -33,15 +33,38 @@ class << self
...
+        def restart(timestamp = nil)
           # TODO: implement using nex_client
           # For now this work with one node
 
           # Restart myself
           # nex_app.restart
           # puts "success" unless nex_app.errors.any?
 
-          FileUtils.touch('tmp/restart.txt')
+          cmd = "touch tmp/restart.txt && " \
```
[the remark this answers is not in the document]
ouranos: It'd also be good to add a sleep between retry :)

api/app/controllers/mno_enterprise/jpi/v1/admin/tenants_controller.rb
ouranos: What do you think of 
```ruby
tenant_params.deep_merge!({frontend_config: {config_timestamp: timestamp}}
```

core/lib/mno_enterprise/platform_adapters/nex_adapter.rb
manu-d: Yep, let's do it sequentially

PR #688, opened 2018-04-17
[MNOE-974] App provisioning workflows

adamaziz15:


ouranos:
Rebased to fix conflicts with #695

PR #689, opened 2018-04-17
[MNOE-1034] WIP Marketplace view to use products

iseessel:


iseessel:
This PR:
1) Removes fetching of all the apps in marketplace#index.
2) Fetches only what we need for the marketplace view. 
3) Methods for product's categories, and tiny_descriptions, for the migration from apps to products.

PR #692, opened 2018-04-24
[MNOE-1021] Add product pricing relationship to admin product#update

iseessel:
Problem: Product Pricing relationship was not being created in the admin panel.
Solution: Whitelist product_pricing_id

Additional Notes:

The relationships that are being created in the dashboard's products_controller#update, currently don't do anything unless we call the json_api_client method #save. Because we are hitting our custom endpoint #modify (which are related to the provisioning workflows), which update the attributes passing in the product_pricing_id works. We could call save on the product, but then we would be making two requests to mno-hub, which is not needed. 

Fully tested locally.


api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/subscriptions_controller.rb
```diff
@@ -102,7 +102,7 @@ def fulfill
   def subscription_update_params
     # custom_data is an arbitrary hash
     # On Rails 5.1 use `permit(custom_data: {})`
-    params.require(:subscription).permit(:start_date, :max_licenses, :custom_data).tap do |whitelisted|
+    params.require(:subscription).permit(:start_date, :max_licenses, :custom_data, :product_pricing_id).tap do |whitelisted|
```
adamaziz15: Should we also whitelist `product_contract_id` here? I'm thinking it's something we should be able to update since the previous code was setting a relationship for it.

iseessel: Yes!

PR #695, opened 2018-04-30
[OPAL-353] Make changes to bluesky adapter backwards compatible

iseessel:


iseessel:
@ouranos @BrunoChauvet Note that this PR is not compatible with my changes on mnoe-angular and mnoe-admin panel (only minor changes required)

PR #697, opened 2018-05-02
Use Redis for cache if present

ouranos:
Adding this to the application config template so it's explicit in the
host project.


ouranos:
Hmm my bad we need the redis gem too.

ouranos:
Updated

PR #713, opened 2018-05-16
[MNOE-1099] ux when no price plans in billing currency

NickenChuggets:


ouranos:
Could we have just the commit with the locales? I don't want to mix it with MNOE-993 and expose the feature if it's not ready yet.

PR #717, opened 2018-05-18
[OPAL-433][MNOE-1128] Add phone to address input

iseessel:


iseessel:
@adamaziz15 Build should be fixed.

api/app/views/mno_enterprise/auth/registrations/_address_form.haml
```diff
@@ -22,3 +22,7 @@
   .col-sm-12
     = text_field_tag "phone", nil, class: "form-control", placeholder: "Phone"
 %br
+.row
+  .col-sm-12
+    =  select_tag "phone_country_code", options_for_select(ISO3166::Country.codes.map { |code| c = ISO3166::Country[code]; ["#{c.alpha2} +#{c.country_code}", code]}, MnoEnterprise.app_country),  class: 'form-control'
+%br
```
adamaziz15: Not sure if it's possible but can we have this formatted like the phone number input for new users? Usually is a small dropdown on the left of the phone number input. It's done in the view for when you go to confirm a new user.

PR #725, opened 2018-05-25
[MNOE-1134] Organization name & description not appearing in Audit Log for subscription

yuritomanek:
@ouranos

### DIFF
--- a/core/app/models/mno_enterprise/audit_event.rb
+++ b/core/app/models/mno_enterprise/audit_event.rb
@@ -20,7 +20,7 @@ module MnoEnterprise
     def format_serialized_details
       AUDIT_LOG_CONFIG.fetch('events', {}).fetch(key, '') % metadata.symbolize_keys
     rescue KeyError => e
-      e.message unless key == 'subscription_update'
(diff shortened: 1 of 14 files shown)


core/app/models/mno_enterprise/audit_event.rb
```diff
@@ -20,7 +20,7 @@ def formatted_details
     def format_serialized_details
       AUDIT_LOG_CONFIG.fetch('events', {}).fetch(key, '') % metadata.symbolize_keys
     rescue KeyError => e
-      e.message
+      e.message unless key == 'subscription_update'
```
ouranos: Why did you need this?

yuritomanek: I did this as it was returning the message in the audit log and it looked messy.

yuritomanek:
The `organisation_id` is being set in the `Subscription#to_audit_event` with the following line:
```
event[:organization_id] = relationships.organization&.dig('data', 'id') if relationships.respond_to?(:organization)
```
I'll need to dig further into it and see why its not being passed correctly.

PR #731, opened 2018-05-30
[MNOE-1126] Restrict access to specific ProductInstances for a given team

MAhsenArif:


api/app/views/mno_enterprise/jpi/v1/teams/_team.json.jbuilder
```diff
@@ -1,4 +1,5 @@
 @all_apps ||= MnoEnterprise::App.all.to_a
+@all_products ||= MnoEnterprise::Product.all.to_a
```
adamaziz15: @yuritomanek @ouranos I'm wondering if it would be better to do this the way it is done for the admin view? Not sure if it would improve performance alot, but wouldn't it be more efficient to include `'product_instances.product'` in the teams controller rather than fetching all products?

PR #743, opened 2018-06-22
[MNOE-1213] Clear staff clients when updating to admin

ouranos:
Backporting [code from 3.3](https://github.com/maestrano/mno-enterprise/blob/master/api/app/controllers/mno_enterprise/jpi/v1/admin/users_controller.rb#L93-L96)

### DIFF
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -165,6 +165,11 @@ module MnoEnterprise
       process_custom_result(result)
     end
 
+    def clear_clients!
+      result = update_clients({data: {attributes: {set: []}}})
+      process_custom_result(result)
+    end
+
     # Find a user using a confirmation token
     def self.find_for_confirmation(confirmation_token)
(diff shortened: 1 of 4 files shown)


core/app/models/mno_enterprise/user.rb
```diff
@@ -165,6 +165,11 @@ def update_clients!(input)
       process_custom_result(result)
     end
 
+    def clear_clients!
+      result = update_clients({data: {attributes: {set: []}}})
```
yuritomanek: I'm curious about this line. You're passing `set` as an attribute but the `update_clients` method only permits `add` or `remove`.

adamaziz15: The `update_clients` method does allow `set` but it only allows it if it is the only attribute. It does not allow it if `add` or `remove` are present.

https://github.com/maestrano/maestrano-hub/blob/develop/app/helpers/api/mnoe/v2/relationship_helper.rb#L19

PR #691, opened 2018-04-23
[MNOE-1014] Members have access to team apps

rheasunshine:
@ouranos

PR #745, opened 2018-06-23
[MNOE-1203][MNOE-1014] Product markups & team apps

adamaziz15:
Update products controller to send organization_id in params when
fetching single product.


adamaziz15:
Includes commits from https://github.com/maestrano/mno-enterprise/pull/691

PR #748, opened 2018-06-28
[OPALSUP-257]Add team member error

jin521:
on the Company/Team page, error happens when adding apps on the team and save, BEFORE adding members to this team. the API call to mno-enterprise Teams controller gets correct response, error happens at render html partial because `instance.app` could be nil


adamaziz15:
Closing in favor of https://github.com/maestrano/mno-enterprise/pull/757

PR #757, opened 2018-07-11
[MNOE-1259] 500 when adding team member

adamaziz15:
- Add guarding to team partial views in the event that apps are not
included in the request

- Remove unnecessary guarding from admin team partial

### DIFF
--- a/api/app/views/mno_enterprise/jpi/v1/admin/teams/_team.json.jbuilder
+++ b/api/app/views/mno_enterprise/jpi/v1/admin/teams/_team.json.jbuilder
@@ -11,6 +11,6 @@ end
 json.app_instances do
   json.array! team.app_instances do |app_instance|
     json.extract! app_instance, :id, :name
-    json.logo app_instance&.app.logo&.to_s
+    json.logo app_instance.app.logo if app_instance.app
   end
 end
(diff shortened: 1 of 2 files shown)


api/app/views/mno_enterprise/jpi/v1/teams/_team.json.jbuilder
```diff
@@ -10,6 +10,7 @@ end
 json.app_instances do
   json.array! team.app_instances do |instance|
     json.extract! instance, :id, :name
-    json.logo instance.app.logo
+    # Check if app is present as it is not needed for create/update actions
+    json.logo instance.app.logo if instance.app
```
iseessel: why are we calling #to_s in the admin jbuilder and not here?

api/app/views/mno_enterprise/jpi/v1/admin/teams/_team.json.jbuilder
```diff
@@ -11,6 +11,6 @@ end
 json.app_instances do
   json.array! team.app_instances do |app_instance|
     json.extract! app_instance, :id, :name
-    json.logo app_instance&.app.logo&.to_s
+    json.logo app_instance.app.logo.to_s if app_instance.app
```
iseessel: Whats wrong with the safe navigation operator?

adamaziz15: Just think it's more explicit to do it this way

api/app/views/mno_enterprise/jpi/v1/teams/_team.json.jbuilder
adamaziz15: Good question, going to remove the `to_s` in the admin view, I don't see why it would be there honestly

PR #760, opened 2018-07-30
[OPAL-532] Added locales for quote

MAhsenArif:


adamaziz15:
@ouranos @MAhsenArif  Will most likely close this one to move these into the opal webstore as they are specific to the opal pricing breakdown

PR #763, opened 2018-08-07
[MNOE-1266] Change of Maestrano name in the content

ruthw:
update sanitized description to not replace for bucket name cdn-prd-maestrano

### DIFF
--- a/core/app/models/mno_enterprise/app.rb
+++ b/core/app/models/mno_enterprise/app.rb
@@ -61,7 +61,7 @@ module MnoEnterprise
     # Sanitize the app description
     # E.g.: replace any mention of Maestrano by the tenant name
     def sanitized_description
-      @sanitized_description ||= (self.description || '').gsub(/(?!cdn\.)maestrano(?!\.com)/i,MnoEnterprise.app_name)
+      @sanitized_description ||= (self.description || '').gsub(/(?<!cdn\.)(?<!cdn-prd-)maestrano(?!\.com)/i,MnoEnterprise.app_name)
     end
 
     # Methods for appinfo flags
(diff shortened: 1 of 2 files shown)


core/app/models/mno_enterprise/app.rb
```diff
@@ -61,7 +61,7 @@ def to_audit_event
     # Sanitize the app description
     # E.g.: replace any mention of Maestrano by the tenant name
     def sanitized_description
-      @sanitized_description ||= (self.description || '').gsub(/(?!cdn\.)maestrano(?!\.com)/i,MnoEnterprise.app_name)
+      @sanitized_description ||= (self.description || '').gsub(/(?<!cdn\.)(?<!cdn-prd-)maestrano(?!\.com)/i,MnoEnterprise.app_name)
```
ruthw: @ouranos, changed the CDN part regex little bit by adding `<`. 

I think the `cdn` part didn't work as it was using `?!cdn`, which is negative lookahead, which won't match `cbn.` before `maestrano`

So by using `?<!cdn` it will match the `cdn` before `maestrano`.

Does this looks right to you? Let me know if I'm misunderstanding anything.

ouranos: This is interesting, it seems that the negative lookahead is working with `cdn.` but not `cdn-prd-`.
So you'd only need negative lookbehind for `cdn-prd-` but let's keep it consistent by using negative lookbehind for both.

ruthw: fail test when remove com check
Actually I don't think negative lookahead works with `cdn`, it worked because we had negative lookahead for `.com`.

As the screenshot above, if I remove negative lookahead for `.com`, and change `cdn` back to negative lookahead by removing the `<`, the test fail.

Is that right?

iseessel: So do we still need the `.com` regex?

ruthw: Good point. Technically we only need one, `cdn.` OR `.com` for the requirement, but with both it also works.

I'm ok with either keep both or only one of them. What do you think @ouranos ?

PR #765, opened 2018-08-09
[MNOE-814] Remove flash messages during OAuth workflow

ruthw:
Update oauth controller to render public layout instead of default application layout, as it doesn't have flash message displayed. 

Attached screenshots to JIRA ticket.

### DIFF
--- a/api/Gemfile
+++ b/api/Gemfile
@@ -1,6 +1,7 @@
 eval(File.read(File.dirname(__FILE__) + '/../common_mnoe_dependencies.rb'))
 
 gem 'mno-enterprise-core', path: '../core'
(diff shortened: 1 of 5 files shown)


api/Gemfile
```diff
@@ -1,6 +1,7 @@
 eval(File.read(File.dirname(__FILE__) + '/../common_mnoe_dependencies.rb'))
 
 gem 'mno-enterprise-core', path: '../core'
+gem 'mno-enterprise-frontend', path: '../frontend'
```
ruthw: The Travis build was failing because the spec in `api` can't find the layout in `frontend`, so had to include `frontend` gem in `api`.

PR #768, opened 2018-08-13
[OPAL-531] Users not added to the Company's member list immediately after adding to the list

MAhsenArif:
Should be tested alongwith OPALSUP-286


ouranos:
With this fix, if an admin (control panel) adds users without sending the invitations they would now display in the customer view, wouldn't they?
Is that an issue?

MAhsenArif:
@ouranos They are already displayed on frontend whenever the user refreshes page. The organization `show` action is rendered and it's view already has all members listing.
[Reference](https://github.com/maestrano/mno-enterprise/blob/4.0/api/app/views/mno_enterprise/jpi/v1/organizations/show.json.jbuilder#L3).

This fix just makes sure that the same listing gets shown on frontend without refresh

PR #773, opened 2018-08-23
[OPALSUP-280] Sort marketplace server-side

jlove-nz:
Provide clients asking for the marketplace apps and products a sorted
list of data. This does away with then need for client side sorting.

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
@@ -81,7 +81,7 @@ module MnoEnterprise::Concerns::Controllers::Jpi::V1::MarketplaceController
 
   # Returns the list of active apps sorted by rank then alphabetically
   def fetch_apps
-    Rails.cache.fetch("marketplace/index-apps-#{@last_modified}-#{I18n.locale}-#{parent_organization_id}") do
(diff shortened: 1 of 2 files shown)


api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb
```diff
@@ -82,14 +82,14 @@ def parent_organization_id
   def fetch_apps
     Rails.cache.fetch("marketplace/index-apps-#{@last_modified}-#{I18n.locale}-#{parent_organization_id}") do
       apps = MnoEnterprise::App.fetch_all(app_relation(parent_organization_id).includes(:app_shared_entities, { app_shared_entities: :shared_entity }).where(active: true))
-      apps.sort_by! { |app| [app.rank ? 0 : 1, app.rank] } # the nil ranks will appear at the end
+      apps.sort_by! { |app| [app.rank ? 0 : 1, app.rank, app.name ? app.name.downcase : ''] } # the nil ranks will appear at the end. Nil names at the start
```
ouranos: If the apps are already sorted alphabetically, sort_by will just re-arrange them by rank, so we can do:
```ruby
apps = MnoEnterprise::App.fetch_all(app_relation(parent_organization_id).includes(:app_shared_entities, { app_shared_entities: :shared_entity }).where(active: true)).order(:name)
apps.sort_by! { |app| [app.rank ? 0 : 1, app.rank] } # the nil ranks will appear at the end
```

PR #774, opened 2018-08-26
[MARLIN-142] Improve connection redirections

alachaum:
Improve redirection workflows on `OAuthController`

*Summary*
- Capture redirect path on `sync` and `disconnect`
- Add option to skip redirection speedbump

### DIFF
--- a/api/lib/mno_enterprise/concerns/controllers/webhook/o_auth_controller.rb
+++ b/api/lib/mno_enterprise/concerns/controllers/webhook/o_auth_controller.rb
@@ -11,10 +11,15 @@ module MnoEnterprise::Concerns::Controllers::Webhook::OAuthController
     before_filter :authenticate_user!, only: [:authorize, :disconnect, :sync]
     before_filter :redirect_to_lounge_if_unconfirmed
     before_filter :check_permissions, only: [:authorize, :disconnect, :sync]
+    before_filter :capture_redirect_path_session, only: [:authorize, :disconnect, :sync]
 
     PROVIDERS_WITH_OPTIONS = ['xero','myob']
     helper_method :main_logo_white_bg # To use in the provision view
     private
+      def capture_redirect_path_session
+        session[:redirect_path] = params[:redirect_path] if params[:redirect_path].present?
+      end
+
       def app_instance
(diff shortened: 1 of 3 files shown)


api/spec/controllers/mno_enterprise/webhook/o_auth_controller_spec.rb
```diff
@@ -47,6 +47,16 @@ module MnoEnterprise
 
       it { subject; expect(response).to be_success }
       it { subject; expect(assigns(:redirect_to)).to eq(redirect_url) }
+      it { subject; expect(session[:redirect_path]).to eq(extra_params[:redirect_path]) }
+
+      context 'when speedbumps must be skipped' do
+        it 'redirects immediately' do
+          ClimateControl.modify SKIP_REDIRECTION_SPEEDBUMPS: 'true' do
```
ouranos: We have a helper around this:
```ruby
with_modified_env SKIP_REDIRECTION_SPEEDBUMPS: 'true' do
end
```

api/lib/mno_enterprise/concerns/controllers/webhook/o_auth_controller.rb
```diff
@@ -57,17 +62,19 @@ def error_message(error_key)
...
     # Certain providers require options to be selected
     if !params[:perform] && app_instance.app && PROVIDERS_WITH_OPTIONS.include?(app_instance.app.nid.to_s)
       render "mno_enterprise/webhook/o_auth/providers/#{app_instance.app.nid}"
       return
     end
 
+    # MnoHub redirection to perform the authorise call
     @redirect_to = MnoEnterprise.router.authorize_oauth_url(params[:id], extra_params.merge(wtk: wtk))
+
+    if ENV['SKIP_REDIRECTION_SPEEDBUMPS']
```
ouranos: Should we use a [feature flag](https://github.com/maestrano/mno-enterprise/blob/5.0/docs/feature-flags.md) instead?

If the Tenant is not allowed to change this, we can have it as `Tenant#metadata` flag

alachaum:
@ouranos Ok I've reworked it to use a feature flag. Displaying a connection speedbump or not should be controlled by the Tenant itself. It's a UX thing after all.

PR #784, opened 2018-09-15
[MNOE-1288] Fix invoices access

ouranos:
## Fix Invoices Access
`user.organizations` is an Array so find is not the JsonAPIClient method but the Enumerable method which returns an enumerator over the array.
The authorisation check was done over the user's first Organization instead of the invoice's Organization.
This was wrongly fixed in #664 

## Move Invoice controller out of JPI
In #664, it was move from `frontend` to `api` but it shouldn't have been moved in JPI as this is not a JPI controller but a normal controller rendering html (well pdf)


ouranos:
Rebased to be on the latest `4.0` version. Going to build a demo with it.

ouranos:
@adamaziz15 could you review and merge? This passed QA.

PR #785, opened 2018-09-16
[MNOE-1287] Delisted apps

adamaziz15:
Update tenant config schema to include only active apps


adamaziz15:
@ouranos Rebased on 4.0

core/app/models/mno_enterprise/tenant_config.rb
```diff
@@ -156,9 +156,9 @@ def self.update_locales_list!
     # Populate the app list after fetching it from MnoHub
     # TODO: replace with a Proc.call
     def self.update_application_list!
-      available_app_nids = App.all.map(&:nid)
+      available_app_nids = App.where(active: true).map(&:nid)
```
ouranos: Should we add a default scope to App?
I don't think mnoe should ever see inactive apps?

ouranos: Or it could be done on the hub side

adamaziz15: Makes sense, maybe best to add it as a filter at the resource level in hub?

adamaziz15:
@ouranos Updated specs, looks like they were failing due to updates in nokogiri or sanitize gems. Build was then failing due to bundler failures in travis caused by outdated rubygems. I've added a temp fix for this until it is fixed in travis (added `gem update --system` to `before_install:` in travis config).

core/app/models/mno_enterprise/tenant_config.rb
ouranos: Yeah, should be by default I reckon?

The inactive apps shouldn't be exposed on the API. We only need them in the DevPlatformSync job to re-enable them if they get re-activated from the DevPlatform

PR #795, opened 2018-11-09
Fix health check

ouranos:
Remove redis-if-present check as it's based on the gem being loaded and
not on REDIS_URL presence. The cache check is sufficient in our case.


ouranos:
I'm gonna add a health check for the platform adapter to verify access to persist assets (local FS or Minio)

ouranos:
@adamaziz15 Added Minio health check

PR #797, opened 2018-11-22
modified the send-confirmation-email and reset-password-email methods…

steven-chang-maestrano:
I was thinking this way, the macaw-enterprise( or whoever ) can define their own custom url as a method and mno-enterprise can use it if detected. 

Let me know what you think, 
had some other ideas but they were tricky to implement

### DIFF
--- a/api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
+++ b/api/lib/mno_enterprise/concerns/mailers/system_notification_mailer.rb
@@ -36,13 +36,19 @@ module MnoEnterprise::Concerns::Mailers::SystemNotificationMailer
   #   :confirmation_link
   #
   def confirmation_instructions(record, token, opts={})
+    if defined?( custom_confirmation_url )
+      confirmation_link = custom_confirmation_url( token )
+    else
+      confirmation_link = user_confirmation_url(confirmation_token: token)
+    end
+
     update_email = record.confirmed? && record.unconfirmed_email.present?
     template = update_email ? 'reconfirmation-instructions' : 'confirmation-instructions'
     email = update_email ? record.unconfirmed_email : record.email
     MnoEnterprise::MailClient.deliver(template,
       default_sender,
       recipient(record).merge(email: email),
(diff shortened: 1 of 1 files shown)


adamaziz15:
@ouranos @steven-chang-maestrano What if we overwrite the helpers in the mailer class? Then we could use the same url helpers, and allow the configuration to happen via env vars or settings. For example, maybe something like this:
```ruby
# In settings.yml
auth:
  confirmation_path: 'auth/register/confirm'

# In system_notification_mailer.rb
def user_confirmation_url(opts = {})
  if Settings.auth.confirmation_path.present?
    "#{default_url_options[:protocol]}://#{default_url_options[:host]}/#{Settings.auth.confirmation_path}?#{opts.to_query}"
  else
    super
  end
end
```

steven-chang-maestrano:
> @ouranos @steven-chang-maestrano What if we overwrite the helpers in the mailer class? Then we could use the same url helpers, and allow the configuration to happen via env vars or settings. For example, maybe something like this:
> 
> ```ruby
> # In settings.yml
> auth:
>   confirmation_path: 'auth/register/confirm'
> 
> # In system_notification_mailer.rb
> def user_confirmation_url(opts = {})
>   if Settings.auth.confirmation_path.present?
>     "#{default_url_options[:protocol]}://#{default_url_options[:host]}/#{Settings.auth.confirmation_path}?#{opts.to_query}"
>   else
>     super
>   end
> end
> ```

@adamaziz15 - It's not coupled to the 'mnoe/auth/users/confirmation' so I think your solution is better! Have you tested it to see if it works? 

I was looking at a solution where we could overwrite the user_confirmation_url method as well but got stuck getting the root path! I like this defaul_url_options idea.

adamaziz15:
@steven-chang-maestrano I tested it a bit locally and it seems to be ok 👍 

My first thought was to do this at a higher level since there are some other urls that will probably need to be configurable (password expiration page for sure), but I was worried about causing any unwanted side-effects with devise. 👮

PR #799, opened 2018-11-26
[MNOE-1307] Failed login attempts

adamaziz15:
Update devise remote authenticatable strategy so that failed login
attempts can be tracked.

### DIFF
--- a/core/lib/devise/strategies/remote_authenticatable.rb
+++ b/core/lib/devise/strategies/remote_authenticatable.rb
@@ -13,7 +13,7 @@ module Devise
         auth_params = params[scope]
         
         # mapping.to is a wrapper over the resource model
-        resource = mapping.to.new
+        resource = mapping.to.where(email: auth_params[:email]).first || mapping.to.new
 
         return fail! unless resource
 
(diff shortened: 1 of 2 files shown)


core/lib/devise/strategies/remote_authenticatable.rb
```diff
@@ -13,7 +13,7 @@ def authenticate!
         auth_params = params[scope]
         
         # mapping.to is a wrapper over the resource model
-        resource = mapping.to.new
+        resource = mapping.to.where(email: auth_params[:email]).first || mapping.to.new
```
cesar-tonnoir:
```suggestion
        resource = mapping.to.find_or_initialize_by(email: auth_params[:email])
```
?

adamaziz15: Can't use `find_or_initialize_by` here since it's not an active record model. The request for the user is being sent to mnohub.

cesar-tonnoir: by all means / disregard then :)

PR #802, opened 2019-01-31
[MACAWSUP-53] Reflect email branding

steven-chang-maestrano:
Macaw wanted branded email subjects (along with other aspects of the email).

### DIFF
--- a/core/spec/lib/mno_enterprise/smtp_client_spec.rb
+++ b/core/spec/lib/mno_enterprise/smtp_client_spec.rb
@@ -2,54 +2,93 @@ require 'rails_helper'
 
 describe MnoEnterprise::SmtpClient do
   let(:client) { described_class.send(:new) }
+  let(:from_object) { { name: 'John', email: 'john.doe@example.com' } }
+  let(:to_object) { { name: 'Joe', email: 'joe.blogg@example.com' } }
+  let(:template_name) { 'reset-password-instructions' }
+  let(:expected_params) do
+    { from: 'John <john.doe@example.com>',
+      to: 'joe.blogg@example.com',
+      template_path: 'system_notifications',
+      template_name: template_name }
+  end
 
   describe '#deliver' do
-    subject do
-      client.deliver(
-        'reset-password-instructions',
-        {name: 'John', email: 'john.doe@example.com'},
-        {name: 'Joe', email: 'joe.blogg@example.com'},
-        attachments: [
-          {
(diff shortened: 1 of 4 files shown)


steven-chang-maestrano:
@ouranos:
- Changed based to 4.0
- specs written but couldn't get it to run on rspec, I get this error: 
Do I need to do something different?

screen shot 2019-02-04 at 2 55 00 pm

ouranos:
@steven-chang-maestrano build failed. To see how to run the specs have a look at `.travis.yml` or at https://github.com/maestrano/mno-enterprise/blob/4.0/docs/testing.md

steven-chang-maestrano:
@ouranos - Tests have passed and changed base to 4.0. Please review and merge when possible. Let me know if any issues. Thanks

core/spec/lib/mno_enterprise/smtp_client_spec.rb
```diff
@@ -2,54 +2,93 @@
 
 describe MnoEnterprise::SmtpClient do
   let(:client) { described_class.send(:new) }
+  let(:from_object) { { name: 'John', email: 'john.doe@example.com' } }
```
ouranos: I'm not sure you need to extract all params to their own variable (also `from_object` is not the best name ;) )

```ruby
let(:opts) { nil }
let(:params) do
  [
    'reset-password-instructions',
    {name: 'John', email: 'john.doe@example.com'},	
    {name: 'Joe', email: 'joe.blogg@example.com'},
    attachments: [	
      {	
        name: 'some-file.pdf',	 
        content: 'the file'
      }
    ],
   opts	 
  ] 
end

subject { client.deliver(*params) }
```

Then your context can be:
```ruby
context 'with a specified subject' do
  let(:opts) { subject: 'user specified subject' }
  it 'uses the specified subject' do
    expected_params[:subject] = user_specified_subject
    ...
  end
end
```

You don't need to explicitly test `it 'sets the subject of the email to the #humanize(template)'` as it's tested by `calls mail with the correct params`

Then you don't need to add all the extra context as it's the default ('opts has not been inputted by user').
Also note the context wording as recommended by betterspecs (with/when).

cesar-tonnoir:
@ouranos I've applied your recommendations and merged with #803 
Can you please tag when you have a chance? Or do you want me to do it?
I'll lock macaw-enterprise to the commit number in the meantime

PR #804, opened 2019-02-21
[MACAWSUP-43] 2fa for admin users

steven-chang-maestrano:
https://maestrano.atlassian.net/browse/MACAWSUP-43

@ouranos | @cesar-tonnoir - I do the quick response code generation here so that the app name and id can be personalised by each enterprise. Generation and validation of otp happens in the back end via attributes and after_save callbacks. 

Hey my travis ci build is failing but I can't figure it out? Have you guys seen this error before? 

**UPDATE: This was because of sqlite3 incompatibility and was fixed recently in a different pull request.**

### DIFF
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -287,9 +287,9 @@ module MnoEnterprise
 
     def requires_otp_for_login?
       if admin?
-        Settings.authentication.two_factor.admin_enabled
+        Settings&.authentication&.two_factor&.admin_enabled
       else
-        Settings.authentication.two_factor.users_enabled
+        Settings&.authentication&.two_factor&.users_enabled
       end
     end
 
--- a/core/app/models/mno_enterprise/user.rb
+++ b/core/app/models/mno_enterprise/user.rb
@@ -1,3 +1,5 @@
+require 'rqrcode_png'
+
 module MnoEnterprise
   class User < BaseResource
     include MnoEnterprise::Concerns::Models::IntercomUser
@@ -49,6 +51,9 @@ module MnoEnterprise
     property :geo_currency, type: :string
     property :failed_attempts
     property :password_changed_at
+    # Two factor
+    property :otp_attempt_successful, type: :boolean
+    property :unconfirmed_otp_secret, type: :string
 
     has_one :sub_tenant
 
@@ -278,6 +283,45 @@ module MnoEnterprise
 
     def admin?
       admin_role.to_s.casecmp(ADMIN_ROLE).zero?
+    end
+
+    def requires_otp_for_login?
+      if admin?
+        Settings.authentication.two_factor.admin_enabled
+      else
+        Settings.authentication.two_factor.users_enabled
+      end
+    end
+
+    def activate_otp
+      self.attributes = {
+        otp_required_for_login: true,
+        unconfirmed_otp_secret: nil
+      }
+      save!
+      reload
+    end
+
+    def set_quick_response_code_in_attributes
+      two_factor_url = 'otpauth://totp/'\
+                       "#{Settings.authentication.two_factor.app_id}:"\
+                       "#{email}?"\
+                       "secret=#{unconfirmed_otp_secret}"\
+                       "&issuer=#{Settings.authentication.two_factor.app_name}"
+      qr_code = RQRCode::QRCode.new(two_factor_url)
+                               .to_img
+                               .resize(240, 240)
+                               .to_data_url
+      attributes['quick_response_code'] = qr_code
+    end
+
+    def validate_and_consume_otp!(otp_attempt)
+      self.attributes = {
+        otp_attempt: otp_attempt,
+        otp_attempt_successful: false
+      }
+      save!
+      reload.otp_attempt_successful
     end
   end
 end
--- a/core/config/initializers/00_tenant_config_schema.rb
+++ b/core/config/initializers/00_tenant_config_schema.rb
@@ -704,6 +704,40 @@ MnoEnterprise::CONFIG_JSON_SCHEMA = {
           }
         },
       }
+    },
+    authentication: {
+      title: 'Authentication feature',
+      description: 'Authentication feature',
+      type: 'object',
+      properties: {
+        two_factor: {
+          title: 'Two factor',
+          description: 'Two factor',
+          type: 'object',
+          properties: {
+            admin_enabled: {
+              type: 'boolean',
+              description: 'Enable for admin',
+              default: false
+            },
+            app_id: {
+              type: 'string',
+              description: 'App id to display on the authenticator',
+              default: ''
+            },
+            app_name: {
+              type: 'string',
+              description: 'App name to display on the authenticator',
+              default: ''
+            },
+            users_enabled: {
+              type: 'boolean',
+              description: 'Enable for users',
+              default: false
+            }
+          }
+        }
+      }
     }
   }
 }.with_indifferent_access.freeze
--- a/core/lib/mno_enterprise/concerns/controllers/auth/sessions_controller.rb
+++ b/core/lib/mno_enterprise/concerns/controllers/auth/sessions_controller.rb
@@ -30,27 +30,26 @@ module MnoEnterprise::Concerns::Controllers::Auth::SessionsController
   
   # POST /resource/sign_in
   def create
-    @user = warden.authenticate!(auth_options)
-    if @user.requires_otp_for_login?
-      @user.activate_otp
-      if @user.unconfirmed_otp_secret.present?
-        @user.set_quick_response_code_in_attributes
-      end
-      return respond_with @user, location: after_sign_in_path_for(@user)
+    self.resource = warden.authenticate!(auth_options)
+    if resource.requires_otp_for_login?
+      resource.activate_otp
+      resource.set_quick_response_code_in_attributes if resource.unconfirmed_otp_secret.present?
+    else
+      sign_in(resource)
     end
-    sign_in(@user)
-    yield @user if block_given?
-    respond_with @user, location: after_sign_in_path_for(@user)
+    yield resource if block_given?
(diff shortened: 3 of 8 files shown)


core/lib/mno_enterprise/concerns/controllers/auth/sessions_controller.rb
```diff
@@ -29,23 +29,41 @@ module ClassMethods
...
-  #   super
-  # end
-  # def create
-  #   self.resource = warden.authenticate!(auth_options)
-  #   set_flash_message(:notice, :signed_in) if is_flashing_format?
-  #   sign_in(resource_name, resource)
-  #   yield resource if block_given?
-  #   respond_with resource, location: after_sign_in_path_for(resource)
-  # end
+  def create
```
ouranos: Backward incompatibility as this:
- no longer yield the resource to a block if provided.
- use @user instead of self.resource

core/lib/mno_enterprise/concerns/controllers/auth/sessions_controller.rb
```diff
@@ -29,23 +29,41 @@ module ClassMethods
...
+    @user = warden.authenticate!(auth_options)
+    if @user.requires_otp_for_login?
+      # This will trigger an after_save callback on the back end
+      # Where if there is an otp_secret then nothing will be changed
+      # But if there isn't, an unconfirmed_otp_secret will be generated.
+      @user = @user.activate_otp
```
ouranos: Should we have a check there to not send the update_attributes request to the backend if not needed?
Or is this also used to generate an `otp`?

core/spec/controllers/mno_enterprise/sessions_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
```
ouranos: I'm confused as what this file is supposed to be testing

core/spec/models/mno_enterprise/user_spec.rb
```diff
@@ -373,6 +372,35 @@ def reload_user
...
+      let(:user) { build(:user) }
+      before(:each) do
+        user.stub(:save!)
+        MnoEnterprise::User.stub(:find) { [user] }
+      end
+
+      it "sets self.attributes['otp_required_for_login'] to true" do
+        user.attributes = {}
+        user.activate_otp
+        expect(user.attributes['otp_required_for_login']).to be true
```
ouranos: Doesn't `user.otp_required_for_login` work? 
Also the description should be more user friendly:
```ruby
it "sets #otp_required_for_login to true" do
  user.attributes = {} # Not sure about this though...
  user.activate_otp
  expect(user.otp_required_for_login).to be true
end
```

core/spec/models/mno_enterprise/user_spec.rb
```diff
@@ -373,6 +372,35 @@ def reload_user
       end
     end
 
+    describe '.activate_otp' do
```
ouranos: You shouldn't be describing the implementation but the behavior

Your spec is currently too coupled to the implementation, if you change it as I suggested (eg: `reload`), it will break the spec although the behavior is still correct.

core/spec/models/mno_enterprise/user_spec.rb
```diff
@@ -385,6 +413,68 @@ def reload_user
       end
     end
 
+    describe '.requires_otp_for_login?' do
+      let(:user) { build(:user) }
+
+      describe '#admin?' do
```
ouranos: - Prefer context to describe the context
- Avoid stubbing too much (we have an [admin trait on the factory](https://github.com/maestrano/mno-enterprise/blob/5.0/core/lib/mno_enterprise/testing_support/factories/users.rb#L46-L48))

This can be rewritten this way

```ruby
describe '.requires_otp_for_login?' do
  subject { user.requires_otp_for_login? }
  
  context "when the user is an admin" do
    let(:user) { build(:user, :admin) }

    context 'when 2FA is required for admin' do
      before { Settings.authentication.two_factor.admin_enabled = true }

      it { is_expected.to be true }
    end

    context 'when 2FA is not required for admin' do
      before { Settings.authentication.two_factor.admin_enabled = false }

      it { is_expected.to be false }
    end
  end
end
```

core/config/initializers/00_tenant_config_schema.rb
```diff
@@ -704,6 +704,40 @@
...
+          type: 'object',
+          properties: {
+            admin_enabled: {
+              type: 'boolean',
+              description: 'Enable for admin',
+              default: false
+            },
+            app_id: {
+              type: 'string',
+              description: 'App id to display on google authenticator',
```
ouranos: -google
+the

core/config/initializers/00_tenant_config_schema.rb
```diff
@@ -704,6 +704,40 @@
...
+      title: 'Authentication feature',
+      description: 'Authentication feature',
+      type: 'object',
+      properties: {
+        two_factor: {
+          title: 'Two factor',
+          description: 'Two factor',
+          type: 'object',
+          properties: {
+            admin_enabled: {
```
ouranos: Might it be better to call this required? as it's effectively required?
Or do we keep it this way to have the possibility to have opt in 2FA later (eg: 2fa is enabled but not required)

core/config/initializers/00_tenant_config_schema.rb
```diff
@@ -704,6 +704,40 @@
...
+              default: ''
+            },
+            app_name: {
+              type: 'string',
+              description: 'App name to display on google authenticator',
+              default: ''
+            },
+            non_admin_enabled: {
+              type: 'boolean',
+              description: 'Enable for non-admin',
```
ouranos: users instead of non-admin?

core/config/initializers/00_tenant_config_schema.rb
steven-chang-maestrano: That's king of what I was thinking but up to you

core/app/models/mno_enterprise/user.rb
```diff
@@ -1,3 +1,5 @@
+require 'rqrcode_png'
```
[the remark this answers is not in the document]
steven-chang-maestrano: @ouranos I get an error if this is not here?

core/spec/models/mno_enterprise/user_spec.rb
steven-chang-maestrano: I think I kind of get you and have modified it accordingly. Let me know what you think.

cesar-tonnoir:
Some comments from Oli haven't been addressed.
Merging latest changes from 5.0

@ouranos FYI

core/lib/mno_enterprise/concerns/controllers/auth/sessions_controller.rb
cesar-tonnoir: @steven-chang-maestrano why is this not addressed after 13 days?
doing it myself

core/config/initializers/00_tenant_config_schema.rb
cesar-tonnoir: @ouranos given your comment I'm unsure what solution you actually prefer, so keeping as is

core/app/models/mno_enterprise/user.rb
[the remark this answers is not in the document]
cesar-tonnoir: seems to be required by method `set_quick_response_code_in_attributes` ?

cesar-tonnoir:
I've tested again in UAT and it seems to work fine - @adamaziz15 can you please review 1 last time?

core/app/models/mno_enterprise/user.rb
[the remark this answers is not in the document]
adamaziz15: It shouldn't be needed, it seems to be an issue related to eager loading

core/app/models/mno_enterprise/user.rb
```diff
@@ -279,5 +284,44 @@ def staff?
     def admin?
       admin_role.to_s.casecmp(ADMIN_ROLE).zero?
     end
+
+    def requires_otp_for_login?
+      if admin?
+        Settings.authentication.two_factor.admin_enabled
+      else
+        Settings.authentication.two_factor.users_enabled
```
adamaziz15: Need guarding here ^^ - just like the routes file

core/app/models/mno_enterprise/user.rb
```diff
@@ -279,5 +284,44 @@ def staff?
...
+      end
+    end
+
+    def activate_otp
+      self.attributes = {
+        otp_required_for_login: true,
+        unconfirmed_otp_secret: nil
+      }
```
adamaziz15: Do we want to just set this here? Or better to merge? Not sure if there will be other attributes that are needed (now or in the future) might be better to merge just in case.

core/app/models/mno_enterprise/user.rb
```diff
@@ -279,5 +284,44 @@ def staff?
...
+                       "&issuer=#{Settings.authentication.two_factor.app_name}"
+      qr_code = RQRCode::QRCode.new(two_factor_url)
+                               .to_img
+                               .resize(240, 240)
+                               .to_data_url
+      attributes['quick_response_code'] = qr_code
+    end
+
+    def validate_and_consume_otp!(otp_attempt)
+      self.attributes = {
```
adamaziz15: same thing about merging here

core/spec/models/mno_enterprise/user_spec.rb
```diff
@@ -385,6 +417,82 @@ def reload_user
       end
     end
 
+    describe '#requires_otp_for_login?' do
+      subject { user.requires_otp_for_login? }
+
+      context 'when user is an admin' do
+        let(:user) { build(:user, :admin) }
```
adamaziz15: not a huge deal, but we can refactor this so it's not necessary to redefine `user` over and over again.

PR #805, opened 2019-03-13
Add dependency constraint on sqlite3

ouranos:
sqlite3 1.4 is conflicting with sqlite3_adapter which requires 1.3.x

See https://github.com/rails/rails/issues/35153


steven-chang-maestrano:
@ouranos - Does it matter that we're using 5.0 branches and this is merging into 4.0 branch?

PR #806, opened 2019-03-13
[MACAW-148] Add Intercom Identity Auth to JPI v2 Users

xaun:


api/spec/controllers/mno_enterprise/jpi/v2/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+          .to_return(status: 200, body: stub_resp.to_json, headers: {})
+      end
+
+      it { is_expected.to have_http_status(:ok) }
+      it { expect(stub).to have_been_requested }
+      it { expect(JSON.parse(response.body)).to eq(stub_resp) }
+
+      context 'when intercom is enabled' do
+        before do
+          allow_any_instance_of(MnoEnterprise::IntercomUserIdentity).to receive(:new).and_return(
```
ouranos: Ah it's because you mixing `allow_any_instance_of` with `new`

So either:
```ruby
allow(MnoEnterprise::IntercomUserIdentity).to receive(:new).and_return(double(....

# OR

allow_any_instance_of(MnoEnterprise::IntercomUserIdentity).to receive(:intercom_enabled?).and_return(true)

```

ouranos:
Actually, couldn't you reuse this: https://github.com/maestrano/mno-enterprise/blob/master/core/lib/mno_enterprise/concerns/models/intercom_user.rb

and have intercom_enabled in the suer controller?

xaun:
@ouranos please re-review

api/app/controllers/mno_enterprise/jpi/v2/users_controller.rb
```diff
@@ -1,8 +1,24 @@
...
+
     def update_password
       resp = MnoHubClient.patch(File.join(endpoint, params.require(:id), 'update_password'), {body: request.raw_post}.merge(authentication_hash))
       render_results(resp)
     end
+
+    private
+
+    def apply_intercom_auth(response)
+      user_hash = current_user.intercom_user_hash
```
ouranos: When Intercom is disable you'll have `no method error intercom_user_hash`?

xaun: @ouranos in 5.0 the `IntercomUser` concern is always included https://github.com/maestrano/mno-enterprise/blob/5.0/core/app/models/mno_enterprise/user.rb#L3. 

Please advise? This change will be merged into 5.0 right? Should I add `MnoEnterprise.intercom_enabled?` guarding to the v2/users_controller?

ouranos: Ok my bad.

Yes better to had the guard then

PR #808, opened 2019-03-28
Code Improvements

ouranos:
Some preliminary changes before the Staff Dashboard PR.
I split this in its own PR to facilitate code review as there seems to be quite a bit of change but it's mostly moving code around.

* Add missing specs from the dashboard template implementation (the error paths were not tested)
* Fix `currency_as_string` deprecation warnings (backported from #783)
* Decorate Admin::OrganizationController to allow customisation in host project (no code changes except the following Rubofix)

Note that `3.3` has been merged in `master` to bring the master branch up to date with the latest feature in `3.3` (dashboard templates)

`master` will be tagged `3.4` after the staff dashboard release

PR #809, opened 2019-03-28
[KAPP-175] Staff Dashboard

ouranos:
## Staff Dashboard
This feature allow staff to create their own dashboards to view their customer data from the admin panel.
The dashboard are owned by the staff but use the customer company as data source.
This require the IDMv2 to be enabled so that staff members are authorised to view their customer data.

See the (incoming) mnoe-admin-panel PR for screeenshots.

Most of the changes boil down to adding a new `/jpi/v1/admin/impac/dashboards` to allow staff to manipulate dashboard.

The commits should be logically isolated to make the review easier.
~I've separated 995547c...2415f52 in #808 as it's just general improvement and might help with the review.~ Rebased

I also had a stab at DRYing the specs (a69662b7):
* JpiV1AdminController context for all the basic setup
* Reduce duplication between Dashboard and DashboardController
Let me know if that actually makes it harder, I'll drop the commit.
Or if you have other ideas.

Know issues:
* Advisor Dashboards don't support KPI yet (not enabled for KAPPA). Should be easy to implement based on the Widget.

### DIFF
--- a/api/app/controllers/mno_enterprise/jpi/v1/admin/impac/dashboards_controller.rb
+++ b/api/app/controllers/mno_enterprise/jpi/v1/admin/impac/dashboards_controller.rb
@@ -71,10 +71,32 @@ module MnoEnterprise
       head status: :ok
     end
 
+    # Allows to create a dashboard using another dashboard as a source
+    # At the moment, only dashboards of type "template" can be copied
+    # Ultimately we could allow the creation of dashboards from any other dashboard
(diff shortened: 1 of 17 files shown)


cesar-tonnoir (approved):
Gosh! that's a PR :)
Looks good to me, but I'll take some more time to have a look in details.

adamaziz15 (approved):

PR #810, opened 2019-04-19
[MACAWSUP-93] Failed login attempt error

adamaziz15:
Use custom endpoint to support devise lockable module.
Devise uses the ActiveRecord ::increment_counter method to record failed
login attempts. The custom endpoint will trigger this method call in
MnoHub.


ouranos (dismissed):
LGTM but I think this should be opened on the `4.0` branch

adamaziz15:
@ouranos Updated the base to 4.0 👍

ouranos (approved):

PR #812, opened 2019-05-09
[MARLIN-748] add deeplink endpoint to pages controller

xiuzhong:
As described on the ticket, *mno-enterprise* need to expose the _deeplink_ endpoint as below:

`get '/deeplink/:oid/:etype/:eid', to: 'pages#deeplink', constraints: {oid: /[\w\-\.]+/, eid: /[\w\-]+/}`

### DIFF
--- a/api/config/routes.rb
+++ b/api/config/routes.rb
@@ -1,7 +1,7 @@
 MnoEnterprise::Engine.routes.draw do
   # Generic routes
   get '/launch/:id', to: 'pages#launch', constraints: {id: /[\w\-\.:]+/}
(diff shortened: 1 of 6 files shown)


api/config/routes.rb
```diff
@@ -1,6 +1,7 @@
 MnoEnterprise::Engine.routes.draw do
   # Generic routes
   get '/launch/:id', to: 'pages#launch', constraints: {id: /[\w\-\.:]+/}
+  get '/deeplink/:oid/:etype/:eid', to: 'pages#deeplink', constraints: {oid: /[\w\-\.]+/, eid: /[\w\-]+/}
```
ouranos: Same here have a more meaningful name for the variables

ouranos (approved):

PR #815, opened 2019-05-23
[MACAWSUP-117] add devise session limitable

steven-chang-maestrano:
jira: https://maestrano.atlassian.net//browse/MACAWSUP-117

There's a corresponding PR for maestrano hub to add unique_session_id to user's table. Please see jira ticket for link.

For this ticket:

Extension Routes
Not needed as per: https://github.com/phatworx/devise_security_extension/blob/master/lib/devise_security_extension/routes.rb

Devise Extension
Require session_limitable model and add module as per:
https://github.com/phatworx/devise_security_extension/blob/b2ee978af7d49f0fb0e7271c6ac074dfb4d39353/lib/devise_security_extension.rb

Session Limitable model
File created as per:
https://github.com/phatworx/devise_security_extension/blob/b2ee978af7d49f0fb0e7271c6ac074dfb4d39353/lib/devise_security_extension/models/session_limitable.rb

Session Limitable hook
File created as per:
https://github.com/phatworx/devise_security_extension/blob/b2ee978af7d49f0fb0e7271c6ac074dfb4d39353/lib/devise_security_extension/models/session_limitable.rb


ouranos (changes requested):
See comment in JIRA

core/app/models/mno_enterprise/user.rb
```diff
@@ -54,6 +54,7 @@ class User < BaseResource
     # Two factor
     property :otp_attempt_successful, type: :boolean
     property :unconfirmed_otp_secret, type: :string
+    property :unique_session_id, type: :string
```
ouranos: I think this should only be added if `session_limitable` is enabled

core/config/initializers/00_tenant_config_schema.rb
```diff
@@ -710,6 +710,18 @@
       description: 'Authentication feature',
       type: 'object',
       properties: {
+        session_limitable: {
+          title: 'Session limitable',
+          description: 'One session usable per account at once',
```
ouranos: The description should be clearer

core/spec/lib/devise/model/session_limitable_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe Devise::Models::SessionLimitable do
```
cesar-tonnoir: @ouranos I've tried many options, including: 
```ruby
RSpec.describe Devise::Models::SessionLimitable do
  include Warden::Test::Helpers
  
  subject { user.reload }

  let(:user) { build(:user, sso_session: 'session_id') }

  before do
    login_as(user, run_callbacks: true)
```

Nothing seems to be triggering Warden's callbacks... Do you have any clue?

ouranos (commented):
Forgot to submit these comments

core/spec/lib/devise/model/session_limitable_spec.rb
ouranos: See https://github.com/maestrano/mno-enterprise/commit/432b629956266b85c5c3f90543622dff96ce463e for some working specs.

I had to:
- reload the user class after updating the settings so that the module is loaded like in [user_spec](https://github.com/maestrano/mno-enterprise/blob/master/core/spec/models/mno_enterprise/user_spec.rb#L5-L11)
- do a request spec so we go through the full rails stack

The request spec is in `api` to have a controller endpoint to request.

If we use `sso_session` we'll most likely won't need `#update_unique_session_id!` (and the associated hook) as we update sso_session in Maestrano-hub
Just need some testing to make sure it’s always properly updated and to store it in the warden scope.

core/app/models/mno_enterprise/user.rb
cesar-tonnoir: I've completely removed that field and am using `sso_session` instead: isn't it what you suggested?

core/spec/lib/devise/model/session_limitable_spec.rb
cesar-tonnoir: thanks!

cesar-tonnoir:
@ouranos this implementation is working (tested on my local machine). I didn't manage to get the spec working. This environment is too complicated in my opinion and the number of hacks and stubs we have to add to achieve simple results like emulating a user signing in is a bit overwhelming :/.

In this example, no matter what I'm trying, `Warden::Manager.after_set_user only: :fetch` is never called from the spec. The version of devise we're using is 2 years old which isn't helping (we cannot use `Devise::Test::ControllerHelpers` for example)

I'm going to stop there with the spec and go on with the frontend implementation.

core/lib/devise/models/session_limitable.rb
```diff
@@ -0,0 +1,14 @@
+require 'devise/hooks/session_limitable'
```
ouranos: We don't need this model anymore?

Or maybe to just define a method:
```ruby
def session_limitable?
  Settings&.authentication&.session_limitable&.enabled
end
```
Then the hook could be
```diff
- if Settings&.authentication&.session_limitable&.enabled
+ if record.respond_to?(:unique_session_id) && record.session_limitable?
```
Not sure if required though

ouranos:
`Devise::Test::ControllerHelpers` is not needed as `DeviseRequestSpecHelper` is essentially the same.

This feature covers the following aspects which are already complex to test themselves:
- Full stack devise testing with hooks
- Multiple concurrent sessions with devise
- Dynamic configuration (Classes reload)
- External API calls 

So when you combine them it gets quite complex indeed. However, only the last 2 are specific to this framework and are not the most complex to test.

Now, for the reason the specs are not working:
- The `:fetch` event is only triggered when fetching the user from an existing session
- `before { login_as }` is creating a new session per example (and triggers the `:authentication` event)

See https://github.com/maestrano/mno-enterprise/commit/2e3cc996769bd55871a95c4c067d1ddc6dfdc5de for a working spec:
- doing two sequential requests in the same example to trigger the `:fetch` event (using explicit `get` as `subject` is lazy-evaluated and would only trigger 1 request)
- less coupling with the warden implementation by changing the `User#sso_session` to simulate another session rather than changing the session_id stored in the warden session.

I had some issues with the cache and could only find [this](https://stackoverflow.com/a/7838450) so I stubbed it as it's not important for this test case.

cesar-tonnoir:
Thanks @ouranos I've pushed your change and it's all good now.

Sending 2 requests is exactly what I was trying to do, but I misunderstood what `login_as` was doing exactly. The cache problem probably got in my way as well...

Anyway, I think it's finally ready to be merged? :)

ouranos (commented):
Ah damn forgot to submit my comments again 😅

core/lib/devise/models/session_limitable.rb
cesar-tonnoir: I think you can't do:
```ruby
devise_modules << :session_limitable if Settings&.authentication&.session_limitable&.enabled
```
if this model isn't defined. Will try again & at least add the `session_limitable?` method you proposed

cesar-tonnoir: @ouranos Nope - doesn't work well... I've tried different things but it looks like devise needs this module to be defined somewhere if we are to load `session_limitable` in devise' modules with a custom hook...
Ok to be merged?

ouranos: I don't think we need the`devise_modules` as it's to inject the module in the model. We only need the hook. See https://github.com/ouranos/mno-enterprise/commit/a91238c7024a11378c73744e3d21a3b8478d44ca

Might need a last rebase/squash before merging.

cesar-tonnoir: All done

core/lib/devise_extension.rb
```diff
@@ -29,6 +30,7 @@ module Controllers
 
 # modules
 Devise.add_module :password_expirable, controller: :password_expirable, model: 'devise/models/password_expirable', route: :password_expired
+Devise.add_module :session_limitable, model: 'devise/models/session_limitable'
```
ouranos: This one needs to be removed too?

Good to rebase and push to QA otherwise

ouranos (dismissed):

core/lib/devise_extension.rb
cesar-tonnoir: Good catch

PR #816, opened 2019-06-11
[KAPP-191] Fix account manager ACL

ouranos:
Fix account manager ACL which wasn't applied on:
- `#index` search
- `#show`

Also fix the `#index` to use `current_user` for scoping rather than a parameters passed by the frontend.

Requires maestrano/maestrano-hub#1211 for the `#show` scoping


api/lib/mno_enterprise/concerns/controllers/jpi/v1/admin/organizations_controller.rb
```diff
@@ -119,6 +128,15 @@ def user_params
     params.require(:user).permit(:email, :name, :surname, :phone)
   end
 
+  # Scope query to the current account manager
+  def account_manager_scope
```
cesar-tonnoir, line 132: Maybe move to a shared concern? Or application_controller?

cesar-tonnoir (approved):

iseessel (approved):
Nice

